PR #22, opened 2015-02-11
Ie 9-10 client-side rendering

matteofigus:


matteofigus:
IE8 is still broken. Will be fixed later.

Issue #13, opened 2015-01-31
oc-client test coverage


matteofigus:
We need to setup some tests in order to start cleaning up oc-client, perhaps setting up karma and jasmine.
- [x] connect sauce-labs for executing tests via karma
- [x] test coverage for the oc-client

matteofigus:
Good enough for now.

PR #45, opened 2015-03-17
karma

matteofigus:


matteofigus:
This is the infrastructure to do #13

PR #58, opened 2015-04-21
ignore failures in dev server when loading modules

andyroyle:
catch them, log them, continue loading


matteofigus:
Thanks for this one!

Issue #36, opened 2015-02-23
auto install/resolve server.js dependencies on dev watcher


matteofigus:

matteofigus:
#60 this fixes this

PR #60, opened 2015-04-27
Auto install modules

andyroyle:
Turned out to be a bit wider scope than I thought, there are two bits to this:

1) Ensuring that modules which are required are listed in package.json

During the package stage, when identifying local requires, we can also identify modules which are required. Once we have this list we can compare it to the package.json file and throw an error if they are missing.

2) Because we know that all modules required by a component are listed in the package.json, we can read this and install any that aren't in the registry's node_modules folder

A by-product of doing (1) means that if we want to check whether a component requires modules that are not allowed by the server, it is easy to do so at publish-time.


matteofigus:
Wow!
Let me have a look at everything and I'll try to merge it today.

matteofigus:
This fixes #36

andyroyle:
:+1:  awesome

PR #75, opened 2015-05-19
Autocomplete experiment

jankowiakmaria:


cli/index.js
```diff
@@ -18,10 +19,28 @@ var dependencies = {
   registry: new Registry()
 };
 
+var complete = omelette('oc <action>');
+
+complete.on('action', function(fragment, word, line){
+  this.reply(Object.getOwnPropertyNames(commands.oc).sort());
```
matteofigus: can you use _.keys(commands.oc) here for consinstency?

matteofigus: Also, can we move the auto-complete stuff to a separate file, so that we can just to something like
```js
var autocomplete = require('./autocomplete'); 
...
autocomplete.setup(commands);
```

PR #80, opened 2015-06-01
Correct hashKey spelling

tpgmartin:


matteofigus:
Thanks Tom, I'll have a look asap.

matteofigus:
So, all looks good, but perhaps I would like a test so that we are sure we don't break this in the future. Wanna try? If you want we can pair a little bit.

tpgmartin:
Sure, afternoon sometime?

matteofigus:
:+1:

PR #81, opened 2015-06-08
Html components discovery #73

### LINKED ISSUE #73: Html components discovery
When accessing the registry main url with a `Content-Type: text/html` would be cool to
- [x] list the component in HTML format
- [x] provide component information and api using the `/{component}/{version}/~info` route
- [x] ideally, provide a simple sandbox to allow hacking on the component and seeing an html preview using client-side rendering

tpgmartin:
This fixes  #73


matteofigus:
Thanks for the PR. We need to fix the tests first. Can you reproduce locally?

PR #84, opened 2015-06-27
[WIP] using __BASE and moving to AirBnb coding style

doron2402:
1. using __BASE for app root directory
2. using the airbnb coding style, when require first require node modules and later local files.

@matteofigus let me know if you like the changes, it was great way for me to get familiar with the code.

DO NOT merge yet I want the test to pass.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -70,6 +70,7 @@
         "before",
         "beforeEach",
         "after",
-        "afterEach"
+        "afterEach",
+        "__BASE"
     ]
 }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,11 +1,13 @@
 'use strict';
 
+global.__BASE = __dirname;
 var commandsParser = require('./grunt-tasks/support/cli-commands-parser');
 var format = require('stringformat');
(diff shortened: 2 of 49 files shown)


matteofigus:
Hi Doron, thanks for the PR. 

1) I am ok for referencing node modules first
2) Problem of using a global `__BASE` variable is that if any other module uses the same global variable, oc wouldn't work as a dependency because the variable is overwritten by someone else so I would like to not use the node.js global scope.
3) You replaced some of `fs.readFileSync` and `fs.readJsonSync` with `require` - this is not going to work. `require` caches the content of a file, we instead want to read the content during every call because those files change during execution. This is one of the reasons why some tests are breaking. 
4) To locally execute the tests, you need to run `grunt test-local`. To apply the settings you put in `test/mocha.opts`, we perhaps should update the mocha [grunt task](https://github.com/opentable/oc/blob/master/grunt-tasks/mochaTest.js).

doron2402:
@matteofigus maybe we should use a global variable name OC then we can use OC.__BASE or OC.ROOT or something

PR #85, opened 2015-07-02
Ruby gems links

matteofigus:
@todd and team: thanks for the amazing work for getting oc integrated with Ruby!
amazing

I added the links on the home. Feel free to add commits to this branch and merge whenever you are happy about it! 

Thanks!
:beers:


todd:
:+1: Looks good. Still a bit of work to do on the Rails integration, but we're getting there.

PR #88, opened 2015-07-09
err message and stack are stored in getters which don't get resolved.…

stevejhiggs:
… Pull them out so we can see more information about what went wrong


matteofigus:
LGTM, thanks

PR #90, opened 2015-07-10
first go at preprocessors

andyroyle:
First go is just using shell commands, e.g.

package.json:
```
{
  "oc": {
    "preprocessors": [
      {
        "name": "babel",
        "command": "babel"
      }
    ] 
  }
}
```

Downside is that it requires the `babel` cli to be installed, and it doesn't support commands that don't stream. The order matters.

`coffeescript` can be done using the following:

```
{
  "oc": {
    "files": {
        "data": "server.coffee"
    },
    "preprocessors": [
      {
        "name": "coffee-script",
        "command": "coffee",
        "args": ["-sc"]
      }
    ] 
  }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "opn": "1.0.1",
     "read": "1.0.5",
     "semver": "4.3.3",
-    "spawn-sync": "^1.0.11",
+    "spawn-sync": "1.0.11",
     "stringformat": "0.0.5",
     "tar.gz": "0.1.1",
     "uglify-js": "2.4.13",
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "opn": "1.0.1",
     "read": "1.0.5",
     "semver": "4.3.3",
+    "spawn-sync": "^1.0.11",
     "stringformat": "0.0.5",
     "tar.gz": "0.1.1",
     "uglify-js": "2.4.13",
--- a/cli/domain/local.js
+++ b/cli/domain/local.js
@@ -12,6 +12,7 @@ var nodeDir = require('node-dir');
 var request = require('../../utils/request');
 var path = require('path');
 var settings = require('../../resources/settings');
+var spawnSync = require('spawn-sync');
 var Targz = require('tar.gz');
 var uglifyJs = require('uglify-js');
 var validator = require('../../registry/domain/validators');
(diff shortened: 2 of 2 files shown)


andyroyle:
It raises the question of whether we

1) continue spawning the cli (yuck)
2) push the responsibility onto the component to run arbitrary code as a preprocessor (e.g. `babel.compile(...)`)
3) explicitly support common things like babel, coffeescript, browserify inside the oc cli

package.json
```diff
@@ -67,6 +67,7 @@
     "opn": "1.0.1",
     "read": "1.0.5",
     "semver": "4.3.3",
+    "spawn-sync": "^1.0.11",
```
matteofigus: Can we use `1.0.11` version here? This is for consistency.

matteofigus:
Very cool. I also really need this :dancer:

stevejhiggs:
Maybe what we really need to do is preprocess the lot using something like webpack. This would let us just plugin webpack plugins to do the transforms and should also allow arbitrary requires rather than only allowing requiring of json files

matteofigus:
The original plan for this was to have something like the prepublish step on npm, so, something very simple, equivalent of having any grunt watcher doing stuff, but somehow hooked to the oc dev watcher.

So, @andyroyle, I like the idea of OC being un-opinionated but having optional functionalities to make your life easier.
1) This is the plan but keeping it simple enough to be something easy to maintain :+1: 
So, you can run a command :+1: , you can specify the order :+1: , but nothing more than that I would say. And if the command fails it's up to you to fix it (for example taking care of installing globally babel before running it).
2) I still want to allow the components' creators to be able to do this especially for advanced functionalities. So, running a grunt watcher or anything else :+1: 
3) That's what I wanted to do in the first place...problem is to keep the cli's dependency tree small. I wanted to integrate babel, for example, then after looking at its dependency tree I changed my mind :( I already have built-in javascript and css bundling and minification and I am kind of regretting it because the cli has too many dependencies :cry: 

@stevejhiggs the idea of not allowing `require` for js on the server-side part of components is actually for simplicity. Allowing it opens so many scenarios to handle in terms of complexity, security and robustness that we just don't have the resources to handle it (yet?). 
Also, enforces simplicity and modularity in the OC way:

* components nesting in case it makes sense from a domain point of view
* moving to a separate service server-side complexity and keeping your server.js just a proxy - which is actually something that shouldn't heavily load (computationally speaking) the registry's runtimes and something where node.js is actually quite decent in terms of scalability. 

In case you want to modularise your code just for the sake of readability, testability etc, I would say that as soon as it stays simple with a preprocessor you should be able to achieve it anyway. Something like:

```js
// standard node.js
// a.js
module.exports = ...
// server.js
var a = require('./a');
...

// Alternative - client-side kind of way
// a.js
...
// b.js
...
// preprocessor 
cat a.js b.js > server.js

What do you think?

andyroyle:
taken a different tactic with [grunt-oc](/opentable/grunt-oc)

PR #91, opened 2015-07-29
Adding 'troubleshooting' section

ajcw:


matteofigus:
This sounds more like a bug that we can solve, not sure I would add it on the main readme. Perhaps I would prefer in any case just a link on the home and then a `troubleshooting.md` inside the `docs` folder?

PR #94, opened 2015-08-02
Support minifying ie8 compatible css files

char1e5:
Found one issue when publishing reservation-widgets. Test page cannot render background image in css files. It turns out a `clean-css` configuration issue during publishing.
Refer to: https://github.com/jakubpawlowicz/clean-css#how-to-set-compatibility-mode.

Now we can use `oc.ie8css = true` to minify ie8 compatible css files.

### DIFF
--- a/cli/domain/local.js
+++ b/cli/domain/local.js
@@ -102,7 +102,7 @@ module.exports = function(){
   };
 
   var missingDependencies = function(requires, component){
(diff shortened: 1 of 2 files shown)


matteofigus:
This is great! Thanks!

matteofigus:
Published on version `0.16.16`.

Issue #109, opened 2015-09-16
OC Registry should deny publish from old CLIs


matteofigus:
Now that CLI sends User Agent containing node and cli version, we should 
* [x] Deny publishing for old versions asking the user to update
* [x] Deny publishing for old versions of node.js asking the user to update

PR #113, opened 2015-09-22
Added preventing old oc versions to publish

jankowiakmaria:
#109


matteofigus:
Generally speaking is fine, I just would like a better error message, that specifies the registry version. Something like `... Upgrade OC CLI running [sudo] npm i -g oc@1.2.X`. So ideally using the latest patch of the registry's minor version.

This is also for the future in order to prevent people from upgrading CLIs to versions that are actually behind the registry.

matteofigus:
Good job btw! :beers:

PR #116, opened 2015-09-28
if creds are specified on the opts object, don't prompt for them

andyroyle:
@matteofigus trying to find a good way of testing this, unit tests for cli-publish are a bit awkward


matteofigus:
Agree, cli-publish and cli-dev are actually awkward, that's the reason why tests are not great, will make them better in the future.
However, LGTM, let's wait it is green then we can add test coverage later.

andyroyle:
@matteofigus turned out to be easier that I thought, see what you think

matteofigus:
:+1:

Issue #106, opened 2015-09-15
Add Active Directory authentication option


matteofigus:
[This](https://www.npmjs.com/package/activedirectory) works perfectly.

```js
var configuration = {
...
  publishAuth: {
    type: 'active-directory',
    options: {
      url: 'ldap://dc.domain.com',
      baseDN: 'dc=domain,dc=com'
    }
  }
};

var registry = new oc.Registry(configuration);
```

- [x] When publishing, user should be prompted to insert email and password in the CLI. Email should then be remembered and saved inside compiled `package.json` - so that we know who published what.
- [x] Authentication should happen on the server, so as soon as the user is asked to prompt username and password, they should be sent to the registry together with the publish request. That said, it would be a good idea to force the user to use HTTPS endpoints (at least) when doing the publish operation.
- [x] We need to figure out if and how this would fit for the grunt plugin.

@andyroyle it would be good to have your opinion about this

andyroyle:
Awesome :+1: 

As far as the grunt plugin goes, I think the simplest first-step is to have the username/password passed on the [options](https://github.com/opentable/oc/blob/master/cli/facade/publish.js#L18) to the publish step. If they are present then it shouldn't prompt (basically skipping [this method](https://github.com/opentable/oc/blob/master/cli/facade/publish.js#L24-L36))

Beyond that we can try to support token-based or key-based methods if we want.

andyroyle:
https://github.com/andyroyle/oc-auth-ldap

PR #119, opened 2015-10-12
first go at an interface for auth plugins

andyroyle:
interface for an auth plugin:

```javascript
{
  validate: function(authConfig){
    // validate the config object
    // also optionally do any setup the scheme needs
    return {
      isValid: false
      message: 'blarg' // when isValid is false this should contain the error message
    };
  },
  middleware: function(authConfig){
    // should return an express middleware function
    return function(req, res, next){ next(); };
}
```

The plugins are indexed by type, so for an authentication `type: 'foo'` looks for a module called `oc-auth-foo` installed alongside `oc`

e.g.

```
myregistry:
  |- oc
  |- oc-auth-foo
```

And the authConfig will access the plugin using `oc-auth-{type}`.

authConfig: 
```json
{
    "type": "foo",
    "bar": "baz"
}
```

Basic Auth is still included by default.


andyroyle:
relates to #106

andyroyle:
goddammit, I always forget to push the branch to the source fork so that the tests don't break

PR #123, opened 2015-10-14
don't set reqinfo on this

andyroyle:
It blows up in cases when [this is undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), and it's not used anywhere. 

If we have need of magic metadata inside plugins then we should revisit this.

### DIFF
--- a/registry/routes/component.js
+++ b/registry/routes/component.js
@@ -143,7 +143,6 @@ module.exports = function(conf, repository){
 
           _.forEach(plugins, function(plugin, pluginName){
             wrapped[pluginName] = function(){
-              this.__reqInfo = requestInfo;
               return plugin.apply(this, arguments);
             };
           });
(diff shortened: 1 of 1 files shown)


andyroyle:
@matteofigus *nudge* if you're happy to merge, i'll do publish

matteofigus:
Just one thing, given that was the only call done to `wrapPlugins`, can you also remove the second argument and cleanup this couple of lines of code? https://github.com/opentable/oc/commit/65a8e5e3e1409415493c1888ef49baf979dcb07a#diff-46910cd31b6b7e605697e6c1db144ce6R161

andyroyle:
do we need to wrap the plugins at all?

matteofigus:
Uhm, you are right, no, we can just undo all of this https://github.com/opentable/oc/commit/80574e835be8e2b0573058a608ba4f5633d124a0

matteofigus:
Thanks.

PR #125, opened 2015-10-21
Prevent infinite loops

### LINKED ISSUE #124: Prevent component code from running infinite loops
Since the components are (currently) executed inside the same process as the registry, a poorly (or maliciously) written component that contains an infinite loop can bring the registry to it's knees (i.e. cause the execution to get stuck inside a never-ending thread).

We could use something like burrito to walk the `server.js` code for a component (either at publish time or execution time) to inject code that can dump out and return control in the event that a loop runs for "a long time".

```javascript
var burrito = require('burrito');

var serverjs = "module.exports.data=function(){ var x=0; while(true){ x = 123; } for(var i = 0; i < 1e10; i++) { console.log('blarg'); } callback(null, {}); };";

var CONST_MAX_ITERATIONS = 1e7;
var monitoredKeywords = ['while', 'for', 'do'];

var src = burrito(serverjs, function (node) {
  if(monitoredKeywords.indexOf(node.name) > -1){
    node.wrap('{ var __ITER = ' + CONST_MAX_ITERATIONS + ';  %s }');
  }

  if(node.parent() && monitoredKeywords.indexOf(node.parent().name) > -1 && node.name === 'block'){
    node.wrap('{ if(__ITER <=0){ throw new Error("loop exceeded maximum allowed iterations"); }  %s __ITER--; }');
  }
});

```

This would wrap the loops in a block and give them an iterator and a check:

```
module.exports.data = function() {
    var x = 0;
    {
        var __ITER = 1e7;
        while (true) {
            if (__ITER <= 0) {
                throw new Error("loop exceeded maximum allowed iterations");
            }
            {
                x = 123;
            }
            __ITER--;
        }
    }
    {
        var __ITER = 1e7;
        for (var i = 0; i < 1e10; i++) {
            if (__ITER <= 0) {
                throw new Error("loop exceeded maximum allowed iterations");
            }
            {
                console.log("blarg");
            }
            __ITER--;
        }
    }
    callback(null, {});
};
```

andyroyle:
closes #124 

This approach doesn't necessarily cope with nested loops, insofar as the `__ITER` variable will get redeclared in the lower scope. 

This means that nested loops will run until the bottom loop hits the limit (not necessarily just the outer). We can guard separately against nested loops if we want to.

### DIFF
--- a/cli/domain/package-server-script.js
+++ b/cli/domain/package-server-script.js
@@ -5,10 +5,32 @@ var format = require('stringformat');
 var fs = require('fs-extra');
 var path = require('path');
 var uglifyJs = require('uglify-js');
+var falafel = require('falafel');
 var _ = require('underscore');
 
 var hashBuilder = require('../../utils/hash-builder');
 var strings = require('../../resources');
+
+var wrapLoops = function(code){
+  var CONST_MAX_ITERATIONS = 1e9; // should this be configurable?
+  var monitoredKeywords = ['WhileStatement', 'ForStatement', 'DoWhileStatement'];
+
+  return falafel(code, function (node) {
+    if(monitoredKeywords.indexOf(node.type) > -1){
+      node.update('{ var __ITER = ' + CONST_MAX_ITERATIONS + ';'
+        + node.source() + '}');
+    }
+
+    if(!node.parent){
+        return;
+    }
+
+    if(monitoredKeywords.indexOf(node.parent.type) > -1 && node.type === 'BlockStatement'){
+      node.update('{ if(__ITER <=0){ throw new Error("loop exceeded maximum allowed iterations"); } '
+        + node.source() + ' __ITER--; }');
+    }
+  }).toString();
+};
 
 var compress = function(code, fileName){
   try {
--- a/cli/domain/package-server-script.js
+++ b/cli/domain/package-server-script.js
@@ -36,11 +36,10 @@ var compress = function(code, fileName){
   try {
     return uglifyJs.minify(code, { fromString: true }).code;
   } catch (e){
-    var m = e.message;
     if(!!e.line && !!e.col){
-      m = format(strings.errors.cli.SERVERJS_PARSING_ERROR, fileName, e.line, e.col, e.message);
+      throw new Error(format(strings.errors.cli.SERVERJS_PARSING_ERROR, fileName, e.line, e.col, e.message));
     }
-    throw m;
+    throw e;
   }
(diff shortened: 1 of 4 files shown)


matteofigus:
This LGTM, but given it is a quite critical functionality, I would actually like if @jankowiakmaria and @pbazydlo could have a look too.

cli/domain/package-server-script.js
```diff
@@ -5,20 +5,41 @@ var format = require('stringformat');
...
 var path = require('path');
 var uglifyJs = require('uglify-js');
+var falafel = require('falafel');
 var _ = require('underscore');
 
 var hashBuilder = require('../../utils/hash-builder');
 var strings = require('../../resources');
 
+var wrapLoops = function(code){
+  var CONST_MAX_ITERATIONS = 1e9; // should this be configurable?
```
matteofigus: Best would probably be to have a default (used during dev) and then get a configurable value on the registry level.

During publishing, CLI would ask the registry for packaging config and then use that value. 
But at the moment, I don't care, this is enough.
Can you just remove the comment and perhaps move the value here? https://github.com/opentable/oc/blob/master/resources/settings.js

test/unit/cli-domain-package-server-script.js
```diff
@@ -36,7 +28,7 @@ var initialise = function(fs, uglifyStub){
         return _.toArray(arguments).join('/');
       }
     }
-  });
+  }, { console: console });
```
matteofigus: can we remove this?

andyroyle:
@matteofigus I've made those changes.

@jankowiakmaria how scary does it look?

jankowiakmaria:
@andyroyle :fearful:

andyroyle:
@matteofigus @jankowiakmaria tell you what, I'll pull the source of all the existing components and run it though the new packager and see if anything shakes loose

matteofigus:
Thanks, that's actually a good idea

andyroyle:
I've run it against a selection of the components in our Preprod registry; Some I couldn't verify because the components don't link to the repo, but so far nothing I've seen worries me.

andyroyle:
@matteofigus so I can't see any immediate problems that will be caused by this.

What do you think? Merge?

matteofigus:
Let me bump the patch for the other PR I just merged, then I would do a merge + minor for this, ok? give me a couple of mins.

matteofigus:
Ok, published in `0.18.0`

PR #130, opened 2015-10-23
add support for dynamic plugin mocks

andyroyle:
The file `cli/domain/get-mocked-plugins.js` isn't covered by any tests at the moment.

Will write some new proper tests for it. Just sending the PR so you can have a look @matteofigus

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -65,10 +65,10 @@ oc dev <dirName> [port]
 
 Parameters:
 
-|Name|Description|
-|----|-----------|
-|dirName|The name of the directory to watch, where the components are stored|
-|port|The port where to start a local oc instance. Default 3000|
+|Name|Description|Choices|
+|----|-----------|-------|
+|dirName|The name of the directory to watch, where the components are stored||
+|port|The port where to start a local oc instance. Default 3000||
 
 ##info
 Shows installed components on the current project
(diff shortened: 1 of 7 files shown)


docs/cli.md
```diff
@@ -138,10 +138,13 @@ Parameters:
...
 |----|-----------|
-|targetType|The type of item to mock|
+|targetType|The type of item to mock (plugin, command)|
 |targetName|The item to mock|
 |targetValue|The mocked value|
 
+example: `oc mock plugin foo bar` adds a static mock for plugin `foo` to return value `bar`.
+example: `oc mock plugin foo ./bar.js` adds a dynamic mock using file `./bar.js`
+
 ##preview
```
matteofigus: The `cli.md` is generated during publish by a grunt task starting from [the cli commands](https://github.com/opentable/oc/blob/master/cli/commands.js) - you need to make this change there in order to get this propagated here - this will make the cli help up-to-date too. ([here](https://github.com/opentable/oc/blob/master/Gruntfile.js#L67-L74) is the grunt task)

matteofigus:
All looks good to me apart from the doc part - thanks for tackling this

docs/cli.md
andyroyle: ah, that explains why the documentation is a bit rubbish! it's generated!

matteofigus: :dart:

Issue #118, opened 2015-10-01
Client-side rendering a component with a blank response blocks the dom


matteofigus:
If the component has even a single space, it is fine.
If multiple components are there, the component shows "Loading..." and prevents other components to be rendered. Gently reported by @ajcw

andyroyle:
it's actually quite a simple issue, turns out it's caused by this line here: https://github.com/opentable/oc/blob/master/components/oc-client/src/oc-client.js#L239

When `data.html` is empty it evaluates to false and so falls into the error handling branch.

It could be mitigated by removing the truthiness check, but maybe this is desirable.

I'm not yet sure why it blocks the remainder of the render, I'll keep digging.

andyroyle:
ah right, so it's because the callback doesn't happen on error: https://github.com/opentable/oc/blob/master/components/oc-client/src/oc-client.js#L240

matteofigus:
Thanks for the investigation, yes agree, we need to fix both problems. I think the first check can just be removed as having a component that returns blank may actually be a common use-case (imagine you want to just return an empty component in case of error or feature toggled off or whatever). So that is not an error.

The un-called callback is definitely something to fix too as that is the cause for the rendering waterfall cycle to be blocked.

PR #132, opened 2015-10-26
have to actually call registry.register for the plugins

andyroyle:
doh!

Once I've got #118 sorted I'll work on improving the tests for `cli/facade/dev.js`


matteofigus:
Published in `0.18.2`

matteofigus:
The fix works :)

PR #133, opened 2015-10-26
Empty component fix

andyroyle:
- Don't treat an empty component as an error
- Always call the callback to avoid blocking the render loop


andyroyle:
closes #118

matteofigus:
Published in `0.18.3`

andyroyle:
:smile:

matteofigus:
Thanks!

Issue #115, opened 2015-09-24
CHANGELOG


matteofigus:
this.

jankowiakmaria:
I've just found https://github.com/skywinder/Github-Changelog-Generator 
list with alternatives: https://github.com/skywinder/Github-Changelog-Generator/wiki/Alternatives

matteofigus:
Nice. I would like to use a node one in order to be binded to the grunt version task. [This](https://github.com/streetlight/grunt-github-changes) looks promising

jankowiakmaria:
@matteofigus as changelog generators use tags to create proper changelog I thought of using https://github.com/rubenv/grunt-git#the-gittag-task as a part of 'version' grunt task. What do you think? (Next step would be to add tags to already released versions and then changelog generation)

PR #134, opened 2015-10-28
Added tagging when upgrading version

jankowiakmaria:
First step of #115

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 Contributing guidelines
 =======================
 
-Contributes are more than welcome! 
+Contributes are more than welcome!
 
 ## Troubleshooting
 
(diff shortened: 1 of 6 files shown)


matteofigus:
This is cool, I actually would like the git push with the version to be another task at this point, that just gets the version automatically?

```sh
grunt version:patch
grunt push
npm publish .
```

jankowiakmaria:
Yep, makes sense - will do.

CONTRIBUTING.md
```diff
@@ -24,9 +24,7 @@ You need to be enabled for doing this.
     * `grunt version:patch` for bugfixes, new features
     * `grunt version:minor` for all breaking changes
     * `grunt version:major` NOT YET. Still need to define milestones for 1.0.0.
-* git add .
-* git commit -m "\<new version\>"
-* git push
+* git push --follow-tags
```
jankowiakmaria: @matteofigus I've updated to above git command (which should:
*Push all the refs that would be pushed without this option, and also push annotated tags in refs/tags that are missing from the remote but are pointing at commit-ish that are reachable from the refs being pushed*) which seems to be more accurate (than pushing twice). 
Don't know if it makes sense now to move it to the grunt task.

matteofigus: ok.

matteofigus:
thanks :+1:

Issue #128, opened 2015-10-23
allow client plugins to depend on (and use) one another


andyroyle:
A-la hapi's `server.dependency(...)`.

matteofigus:
:+1:

andyroyle:
@matteofigus got an implementation for this working, wanted to get your opinion on how to pass the dependencies into the plugins (for the plugins to use).

Registering a dep is done like this `module.exports.dependencies = ['myplugin1', 'myplugin2'];`

So that the plugin can use it's dependencies inside the register function, we need to pass them in. My gut feeling is that we should change the signature for the register function to this:

`module.exports.register = function(options, deps, next){`

where `deps` is an object the similar to `context.plugins` inside server.js. 

Unfortunately this is a breaking change.

We could work around it by setting a property on the options object if you prefer.

matteofigus:
All good for me. If we make deps optional not necessarily it is a breaking change right?
```js
...
register = function(options, deps, next){
if(typeof(deps) === 'function'){
  next = deps;
  deps = [];
}
...
```

andyroyle:
the problem is that the register function is in client-code, so unless they've got that snippet in all their plugins then it'll break.

unless we try and wrap the register function and do some magic. But that might be a bit nuts

matteofigus:
Yes, but we have semver for that (so if you implement plugins on top of a oc-registry you are safe and that shouldn't change anything on components) + I wrote this sentence on the very top of the README: `Disclaimer: This project is still under heavy development and the API is likely to change at any time. In case you would find any issues, check the troubleshooting page.` :dancers:

matteofigus:
PS - I am trying to push to get a v1 out. In this perspective, I am trying to get the minimal requirements to start doing semver properly and then push majors for changes like this. If you wish to share your voice: https://github.com/opentable/oc/milestones/v1.0.0

andyroyle:
:+1:

andyroyle:
published in 0.20

PR #141, opened 2015-11-06
allow plugins to depend on one another (breaking change)

andyroyle:
- deps must be static and exposed as an array of plugin names, e.g. `module.exports.dependencies = ['myplugin', 'myotherplugin']`
- will detect circular dependencies and missing plugins
- registers each plugin only when all it's dependencies have been registered
- recursively registers in cases of deep dependency chains
- passes dependencies as an object to the register method (e.g. use `deps.myplugin()` to execute the dep)

Adds an extra module dep [madge](https://github.com/pahen/madge), which seems pretty heavy (for the tiny piece we're using (checking for circular dependencies).

Not sure if it's worth finding an alternative or just pulling the logic out and wrapping it in a new package.

See what you think @matteofigus.


matteofigus:
:+1: Thanks good job! I would leave the madge optimisation in the future in case.
Just give me some time to look at the PR - it's a lot of code.

PS
madge depends on coffeescript because of Gruntfile.coffee
ok

andyroyle:
actually, fuck madge, I'm going to swap it out for this: https://www.npmjs.com/package/dependency-graph

matteofigus:
`dependencies": {}` :beers:

andyroyle:
closes #128

andyroyle:
@matteofigus you had a chance to look at this one?

matteofigus:
Published `0.20.0`

Issue #136, opened 2015-11-04
Client-side retry mechanism optimisations


matteofigus:
It would be cool for the client-side client to be able to:
* [ ] Optionally set a maximum number of retries
* [ ] Optionally set a parameter to report the retry number (http://server/component/...?__oc_Retry=2)

andyroyle:
exponential backoff as well?

matteofigus:
That would be cool. ATM I think retries happen just every 5 seconds (and the number is configurable)

PR #144, opened 2015-11-13
Oc client limited retries

pbazydlo:
Resolves: https://github.com/opentable/oc/issues/136
Doesn't do exponential backoff.

* Please verify if configuration params make sense.
* When number of retries is exceeded nice error is shown and client stops retrying.
* Retries are counted per component href.


pbazydlo:
in `test/front-end/ocTestConfig.js` i set `oc.conf.retryInterval = 1` so that retries happen very fast and can be reasonably tested, this is hooked up  in `test/configuration/karma-local.js` so that the config is set before oc-client is loaded. it worked when running `grunt test` and `grunt karma:local` but it seems i missed something

test/configuration/karma-common.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
```
pbazydlo: i pulled common configuration from karma-local and karma-sauce into a separate file

test/front-end/ocTestConfig.js
```diff
@@ -0,0 +1,3 @@
+var oc = oc || {};
+oc.conf = oc.conf || {};
+oc.conf.retryInterval = 1;
```
pbazydlo: this is needed so i can test number of retries being made before test timeout

pbazydlo:
@matteofigus passing now, as for running karma:sauce locally we could add `karma:sauce-local` task and install chrome driver, then we would be able to run it locally

test/front-end/ocTestConfig.js
matteofigus: That's ok, it is exactly how parameters should be set

andyroyle:
looks good to me

andyroyle:
@matteofigus want me to publish?

matteofigus:
@andyroyle that would be cool. When in master, can you also add @pbazydlo here? https://github.com/opentable/oc/blob/master/README.md#contributors

pbazydlo:
yay

andyroyle:
published in 0.20.1

matteofigus:
Thanks :+1:

PR #147, opened 2015-11-19
Refactor cli/domain/local

andyroyle:
pull bits into separate files and improve tests


matteofigus:
This is awesome, thanks!

Only question is that `oc.json` on the root, that looks to me is there for testing purposes? In that case, can we move it to the tests folder?

andyroyle:
Ah, I think that just got committed by mistake. I've removed it

matteofigus:
Published `0.19.5`

Issue #138, opened 2015-11-04
gzip


matteofigus:
So it would be cool to have all js and css gzipped by default. In case, we may want to have a setting to disable it on the component level? (not sure it is needed).

Gzip support: http://schroepl.net/projekte/mod_gzip/browser.htm

andyroyle:
I guess this could be as simple as including the compression module: https://github.com/expressjs/compression

matteofigus:
Well I want gzip also on compiled views and css and js files during packaging. We already use `tar.gz` as dependency to do other stuff, if we gzip the stuff before uploading to cdn too then this should be a good win

andyroyle:
So what's the idea for the CSS/JS. If we gzip it and throw it onto the CDN during publish, how do we reference it in the browser? 

How do we tell the CDN to transfer the content with the correct type?

matteofigus:
So, in theory we reference it in the same way and we publish it in the same way using the usual `Content-Type` header (depending on file's type). Then, we need to ensure content comes back from CDN with an additional `Accept-Encoding: gzip` header (the browser should handle it). It [doesn't look like a tricky thing to do](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html#CompressedS3) as soon as we define a standard and simple way to do it.

andyroyle:
"You also need to develop your application to intercept viewer requests and change the request URL based on whether the request includes an Accept-Encoding: gzip header."
...
"Your application determines that Accept-Encoding: gzip appears in the request header and changes the CloudFront URL to request the compressed version of the file."

For requesting the template, that would probably be OK, the registry can just say 'if the client supports gzip then add the .gz extension to the template path'. But for static css/js assets (referenced from template.js) then it's going to be nasty.

matteofigus:
Closing this as is solved.

PR #153, opened 2015-11-23
js.gz + css.gz handling

matteofigus:
This should be the first part of #138 

Basically, when static folder contains files with `.css.gz` or `js.gz` extensions, 
* oc dev watcher handles it (forwards to the client a `Content-Encoding:gzip` header for given resource)
* s3 publish should save the ContentEncoding property to the S3 object (so that Cloudfront should recognise it and forward it)


matteofigus:
Can you review @andyroyle ?

andyroyle:
makes sense to me

andyroyle:
published 0.20.5

matteofigus:
:+1:

PR #154, opened 2015-11-24
Gzip static files on package

andyroyle:
As per #153 (and #138), this code will identify static assets (css, js, template) when packaging and create gzipped copies alongside.

Just thought I'd send the PR because it complements #153 quite nicely.

Maybe we want to disable the behaviour by default for now (control it from a switch in the package.json). See what you think.

### DIFF
--- a/cli/domain/package-static-files.js
+++ b/cli/domain/package-static-files.js
@@ -8,6 +8,7 @@ var nodeDir = require('node-dir');
 var path = require('path');
 var uglifyJs = require('uglify-js');
 var _ = require('underscore');
+var gzip = require('../../utils/gzip');
 
 var strings = require('../../resources');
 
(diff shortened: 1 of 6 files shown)


matteofigus:
In regards of the setting, we have a `minify` parameter that you can set on the component level to minify js/css - that defaults to true.
Maybe we can use the same setting to include/exclude gzip too?

andyroyle:
is it wise to reuse that flag?

I noticed that the CSS component has it set to false: https://github.com/opentable/SiteResourceAPI-nodejs/blob/master/oc-components/css/package.json#L9

Is this for a particular reason?

matteofigus:
Ok, so, let's use another setting, you're right, perhaps `gzip` or `compress` ;)  

Thanks

matteofigus:
@andyroyle turns out that using a decent CDN (just switched all to Akamai here) all of this is not required as the cdn takes care of responding (or not) with gzipped resources automagically when file matches css, js or whatever extension.

I would keep the other old feature (when .gz, file is saved to S3 with Content-Encoding type set to gzip so that browser understands it as gzip if using directly S3 or dumb CDN) but I would probably remove this as not very needed (and still achievable with a custom task inside component).

What do you think?

The main problem we got with "just serving all gzipped" is that we found out that gzip is supported by all browsers, but some "security" browser plugins remove the Accept-Encoding header so that user doesn't get the advantages of compression. Result is that if you send anyway a gzipped resource, browser doesn't execute JS :(

andyroyle:
I figured this might be nullified by using a decent CDN ;)

matteofigus:
Ok so I am closing this. Thanks

PR #159, opened 2015-12-22
should have an explicit dep on phantom

andyroyle:
when installing with npm 3 (or higher) it doesn't resolve the peer-dependency from karma-phantomjs-launcher so we need to specify it explicitly.


matteofigus:
:+1: thanks

PR #164, opened 2016-01-12
New component-retrieved event

matteofigus:
This is needed for having insight about each component, especially when consuming the registry via the new Batch POST endpoint: #163


matteofigus:
- I made some cleanup on the code, just for consistency
- The event handler is now a singleton module - this allows it to be just required in multiple places for firing events without inject the instance on all the dependent classes, still keeping all the subscriptions. Tests attached
- The request event contains the body - useful for the new POST endpoint
- The component-retrieved event is fired and contains information about the component. Tests attached.

@jankowiakmaria can you review?

I apology if it is a big PR containing many things - should have done multiple ones.
Thanks

jankowiakmaria:
LGTM - should I merge (the title of PR suggests that not yet)? @matteofigus

matteofigus:
Yes, sorry

PR #178, opened 2016-01-27
node.js client should accept serverRendering and clientRendering endpoints instead of generic array (breaking change)

matteofigus:
There are 3 main breaking changes here.

* `new Client()` is not accepted anymore. Options need to be passed in during initialisation: `new Client({ //options });`
* follows that `client.config` is deprecated and made private and not accessible anymore
* When initialising with the options, { registries: []} as array is deprecated in favour of an object that contains 0, 1 or both keys of `clientRendering` and `serverRendering`. So, when doing server-side rendering with failover, failover request will be able to have client-side rendering friendly url.


andyroyle:
lol, it's trying to lint the minified files

PR #186, opened 2016-02-11
Preserve language when doing client-side failover

matteofigus:
When making server-side request with a client, if a failover mechanism is in place the url won't preserve the original `accept-language` header. 

With this option (needed only if components depend on Accept-Language) the client adds a `__ocAcceptLanguage` parameter with the original Accept-Language value. The framework, when receiving that parameter, will ignore the Accept-Language header and will use that parameter instead.


matteofigus:
Anyone could do a review? Perhaps @antwhite or @lukasz-lysik or @andyroyle?

antwhite:
@matteofigus LGTM (it works for my components)

matteofigus:
Thanks @antwhite,  can you merge? I'll publish. Thanks

PR #200, opened 2016-03-10
Have optional callbacks in CLI functions

matteofigus:
This allows facades to be used outside of CLI context, for example to be wrapped inside grunt and gulp plugins. This is required in order to feature complete the `grunt-oc` plugin.


jankowiakmaria:
lgtm, merging

PR #204, opened 2016-03-11
add "getRepositoryUrl" in order to support repository package.json property as object

mattiaerre:
# Description

As per the npm documentation https://docs.npmjs.com/files/package.json#repository the **repository** property inside the package.json file should be an object like the following:

``
{
  "type": "git",
  "url": "https://github.com/npm/npm.git"
}
``

I added a **getRepositoryUrl** method against the component object that returns:

- the content of the **url** property if repository is of type object
- the content of the **repository** property itself if repository is of type string (for backwards compatibility)
- **null** otherwise

then in the view (the partial .jade file) the above method is used instead of the property itself.

### DIFF
--- a/registry/routes/component-info.js
+++ b/registry/routes/component-info.js
@@ -29,7 +29,17 @@ module.exports = function(repository){
             return param.example;
           });
         }
-                
+
+        component.getRepositoryUrl = function() {
+          if (typeof this.repository === 'object') {
+            if (this.repository.url)
+              return this.repository.url;
+          }
+          if (typeof this.repository === 'string')
+            return this.repository;
+          return null;
+        }
+
         return res.render('component-info', {
           component: component,
           dependencies: _.keys(component.dependencies),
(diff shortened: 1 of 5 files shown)


registry/routes/component-info.js
```diff
@@ -29,7 +29,19 @@ module.exports = function(repository){
             return param.example;
           });
         }
-                
+
+        component.getRepositoryUrl = function() {
+          if (typeof this.repository === 'object') {
```
matteofigus: can we just use underscore's _.isObject for consistency?

registry/routes/component-info.js
```diff
@@ -29,7 +29,19 @@ module.exports = function(repository){
...
+            if (this.repository.url) {
+              return this.repository.url;
+            }
+          }
+          if (typeof this.repository === 'string') {
```
matteofigus: and _.isString(...) here

matteofigus:
Grazie, looks good.

Would you mind adding those 2 little adjustments + [adding yourself as contributor](https://github.com/opentable/oc/blob/master/README.md#contacts)?

matteofigus:
Thanks good stuff!

PR #208, opened 2016-03-15
When data is undefined, registry should 500

matteofigus:
This happens when a component calls a callback without an error, but also without the data object. 
At the moment, in some scenarios, the registry just times-out.

When a view doesn't require a view-model, either an empty object should be called-back as result, or a component should have an empty data field in its `package.json` (to declare it doesn't require any logic to work).

Example:
```js
module.exports.data = function(cx, cb){
  var data;
  cb(null);
};
```

### DIFF
--- a/registry/routes/helpers/get-component.js
+++ b/registry/routes/helpers/get-component.js
@@ -99,7 +99,8 @@ module.exports = function(conf, repository){
         if(componentCallbackDone){ return; }
         componentCallbackDone = true;
 
-        if(!!err){
+        if(!!err || !data){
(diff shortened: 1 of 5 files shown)


matteofigus:
Can you review this today @mattiaerre or @pbazydlo?

mattiaerre:
Hi @matteofigus I am looking at this now ... bear w/ me :wink:

pbazydlo:
i'm leaving it to you then @mattiaerre

mattiaerre:
LGTM

// cc @matteofigus @pbazydlo

PR #209, opened 2016-03-15
Exit 1 in case of CLI error

### LINKED ISSUE #174: When CLI errors should exit with non 0 code
As discussed with @seif and @antwhite.

It should be just a matter of closing the app with a `process.exit(1)` in the [CLI's facades](https://github.com/opentable/oc/tree/master/cli/facade)

matteofigus:
Fixes #174 

Now that all the cli's facades have an optional callback (not used by the cli) - it is possible, in case it is not provided, to wrap it with a utility that exits with a `process.exit(1)` when a callback is fired with an error.

To test locally with unix/mac: 
* checkout the branch locally
* run a command that should fail like `node oc-cli.js registry add 23456`
* run `echo $?` to see previous command's exit code


matteofigus:
Can you review this @antwhite or @jankowiakmaria?

pbazydlo:
lgtm, it's a scary one though as you're now forcing death instead of silence, won't it break anything?

pbazydlo:
(outside of oc itself)

matteofigus:
@pbazydlo that's what we want if the callback is fired with an error. Previously, it was exiting with 0 in case of both success and fail.

PR #210, opened 2016-03-15
github api request-limit fix

matteofigus:
When generating the new changelog, you need to be logged in into github or most likely you will hit the api rate limit. This should do, while we integrate something better into the grunt pipeline.

This PR also includes the code for publishing the node.js client.

/cc @pbazydlo @jankowiakmaria @andyroyle


matteofigus:
Can anyone review @jankowiakmaria @pbazydlo ? Thanks!

PR #211, opened 2016-03-21
Forward request headers to component

antwhite:


matteofigus:
:+1: this is great, thanks @antwhite

PR #212, opened 2016-03-22
Changelog using git history instead of github api

matteofigus:
This is a replacement for the current Changelog generator. I spent a couple of hours trying all of them, they all seem over complicated for what we want to achieve here. Also, using github api is crazy as hundreds of requests are needed (repo details, one for each commit, one for each tag), and all we need is instead available using git.

This works. Future improvement is to separate this to its own grunt module, but seems good enough for now.

Can you review @pbazydlo or @jankowiakmaria ?


pbazydlo:
lgtm, @matteofigus have you thought about putting that grunt task into public npm module?

matteofigus:
lol, :+1: 
> Future improvement is to separate this to its own grunt module

PR #214, opened 2016-03-22
S3 validation

### LINKED ISSUE #162: When s3 config is wrong, registry shows components_cache_empty
Instead of showing something about s3 connectivity problems, it shows something like:

Connecting to library: open-components/components
components_cache_empty

It would be better to have a more appropriate error message.

matteofigus:
Fixes #162


test/unit/registry-domain-validator.js
```diff
@@ -129,10 +135,98 @@ describe('registry : domain : validator', function(){
...
+            key: 's3-key', region: 'us-west2', secret: 's3-secret'
+          }};
+
+          it('should not be valid', function(){
+            expect(validate(conf).isValid).to.be.false;
+            expect(validate(conf).message).to.equal(errorMessage);
+          });
+        });
+
+        describe('when s3 setting is missing bucket', function(){
```
jankowiakmaria: @matteofigus isn't this test the same as the one above?

matteofigus: You're right. Fixed it.

jankowiakmaria:
:+1: 
btw @matteofigus do you mind if we start using `rebase` instead of `merge`? It makes history much cleaner (no commits like `Merge branch 'master' of...`).

PR #228, opened 2016-04-07
Global params on POST request

### LINKED ISSUE #223: POST endpoint should allow global parameters
Current behaviour for making a request is making a POST to the registry's base url with a payload like
```json
{
  "components": [
    { "name": "component", "parameters": { "domain": "it" }},
    { "name": "component2", "parameters": { "domain": "it", "hello": "world" }}]
}
```
We want, when multiple component requires the same parameters, to be able to pass this parameters globally. Example (The following example aims to be equivalent to the previous one):
```json
{
  "parameters": { "domain": "it" },
  "components": [
    { "name": "component"},
    { "name": "component2", "parameters": { "hello": "world" }}]
}
```

Acceptance:
- [ ] when a global parameter is specified, it is used by all the components no exclusion
- [ ] when a global parameter is specified and component includes the same parameter with different value, the global parameter is overwritten for that component
- [ ] when both global and component parameters are specified, both are going to be used (see example for `hello` parameter)

matteofigus:
Fixes #223 

Can you review @jankowiakmaria @lukasz-lysik ?


jankowiakmaria:
will do @matteofigus

jankowiakmaria:
lgtm, merging

PR #229, opened 2016-04-11
Empty string should be a valid string parameter

matteofigus:
This fixes the problem for both optional and mandatory string parameters for when an empty string is provided both via a GET or a POST api request.

Specifically, currently, when a `a` parameter is an optional string, `...?c=d` is valid, `...?a=&c=d` is not. It should be instead.

Same when mandatory: `...?c=d` errors and still should, but `...?a=&c=4` errors too but it should be valid instead.


federicomaffei:
seems good to me - happy for this to be merged @matteofigus ?

PR #235, opened 2016-04-29
Moved things around

matteofigus:
This is part of an effort for making the github's homepage cleaner and neat.

No code changes here apart from moving files and updating urls.


mattiaerre:
FWIW 😅  LGTM 👍

PR #239, opened 2016-05-05
Explain better what OC is

matteofigus:
This is part of an effort I want to make to explain to people that visit our github repo to better understand what OC is.

The goal is to test different variants of the readme and test how many stars more we can get.

In theory, if we can better explain what OC is in a couple of sentences, we should be able to get more visibility, credibility and interest which ultimately may convert on getting more feedback and contributions.

It would be cool to know what you think, @antwhite @ajcw @lukasz-lysik @mattiaerre @pbazydlo @andyroyle @ryantomlinson @jankowiakmaria (feel free to invite anyone in the conversation).

In this first test:
1) I moved the tests on the bottom
2) Tried to explain more practically what OC is in a couple of steps
3) I used some trendy words like micro-services and React
4) I mentioned that OC is been in prod for a long time

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,20 @@
 ![oc](https://raw.github.com/opentable/oc/master/docs/logo.png)
 =============
 
-OpenComponents is a framework to develop and deploy robust and distributed html components.
+OpenComponents, **microservices in the front-end world**.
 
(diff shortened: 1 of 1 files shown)


andyroyle:
makes sense to me 👍

antwhite:
LGTM 👍

jankowiakmaria:
👍

PR #240, opened 2016-05-05
add some basic example components

andyroyle:
it's always good to have code samples to help people get started!

Adds two basic components to get to grips with some basic concepts (parameters, server-side operations etc)


matteofigus:
Thanks, very good! Will merge when green

PR #241, opened 2016-05-05
adds dependencies badge in readme

a0viedo:


matteofigus:
Hi, and welcome :)

This is great - we definitely need to update a couple of deps! Good spot

PR #243, opened 2016-05-05
[careful] handlebars 3+4

matteofigus:
This is not easy. Handlebars 3 has some security issues (#252).

Unfortunately upgrading to v4 is not a piece of cake, as compiled templates with v3, are not going to be compatible with parser v4. This means that published v3 components are not going to be executable anymore with v4 and they need a new compilation phase.

This means that for a graceful migration, we will
* Introduce Handlebars 4 in this PR as default compiler for new components. This will include the Handlebars 3 polyfill so that pre-compiled templates with previous version will keep working.
* Publish oc with a new minor version so that all the components in a registry will be republished after being recompiled with H4
* Work on a following PR to remove Handlebars 3 polyfill support as a cleaning step + new oc minor publish.


matteofigus:
I think this is ready to be reviewed. This is pretty huge and kind of a big deal so I'd like a LGTM from multiple people, perhaps @antwhite @jankowiakmaria @mattiaerre 

The behaviour here is:
* Handlebars v4 is from here used by the CLI for compiling new components.
* When component is already published with v3 compiler, the client-side client, the node.js client, and the registry all are able to still use v3 for the execution

I tried to add some tests there and there to ensure both versions are supported, but another pair of eyes is good to ensure all is good and covered.

Issue #217, opened 2016-03-23
when container=true, hrefs should be public ones


matteofigus:
As reported by @antwhite, both the node client + api for rendered components should include hrefs using `clientRendering` as base href.

PR #244, opened 2016-05-19
Docs updated

matteofigus:
I tried to think about #217 - and given we fixed this using the public endpoints as baseUrls, I decided that this is quite enough.

BaseUrl usage is mostly for rendering the oc-component href urls and nest components for client-side rendering, so there is no value on using any private endpoint with that. 

This PR is just to clarify this in the docs. Can you review @antwhite ?


matteofigus:
Automerging

PR #247, opened 2016-05-23
#226 Make AMD compliant 

Hobbit71:
Have converted to be AMD and CommonJS compliant. Still needs tests to be written.

PR #248, opened 2016-05-23
Amd

matteofigus:
This PR follows #247 and it is required in order to make sauce labs to run to all platforms /cc @mattiaerre


matteofigus:
All is green after running in all browers, so merging this @Hobbit71

PR #251, opened 2016-06-20
check is dir with lstatSync

### LINKED ISSUE #250: oc dev . fails if file w/o extension found
as per the following screen grab when running `oc dev .` locally if there is a file w/o extension in the folder a 500 error occurs:

screen shot 2016-06-20 at 17 35 05

the problem can be found here:

https://github.com/opentable/oc/blob/master/src/registry/domain/repository.js#L37

mattiaerre:
this will fix #250


matteofigus:
Thanks!

PR #253, opened 2016-06-23
Updated instructions of how to add static resource to the file

jankowiakmaria:
I've renamed the static resource dir to avoid confusion.

#airportCoding

### DIFF
--- a/docs/advanced-operations.md
+++ b/docs/advanced-operations.md
@@ -108,7 +108,7 @@ Context aggregates request data, and the registry's context. It consists of the 
 
 # Add static resource to the component
 
-In this example an image (`static/static_resource.png`) will be our static resource.
+In this example an image (`src/static_resource.png`) will be our static resource.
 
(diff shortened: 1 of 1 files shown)


matteofigus:
✈️

mattiaerre:
hi @jankowiakmaria and @matteofigus 
I was thinking that while you are "dealing" w/ this it would be nice to have the static folder in the example named as "public" rather than "src" which, in my opinion, looks like more the folder containing all the source code of a component. The name "public" on the other hand, is used by other web application frameworks such as [express](http://expressjs.com/en/starter/generator.html) in order to expose static assets such as javascript and css files.
what do you think?
thanks,
-Mattia

jankowiakmaria:
Makes sense for me 👍

mattiaerre:
thanks @matteofigus and @jankowiakmaria please reopen your branch and change the name then 😉 
-Mattia

PR #264, opened 2016-07-05
Docs cleanup

matteofigus:
* Docs moved to wiki
* Logo image included inside the web interface
* Removed CLI docs generation from build task (that is pretty consolidated now - but still left the autogen task in case we want to update the wiki in the future)
* Wiki is already in place

Can you have a look @jankowiakmaria or @mattiaerre?


jankowiakmaria:
in general LGTM

According to wiki changes:
* few page titles don't match titles from index (example: "Registry post route" vs "Batch endpoint") <- is it desired?
* https://github.com/opentable/oc/wiki#components link doesn't work for me (it doesn't scroll to appropriate section of the page)
* not sure if logo in the index is not too big:

image


Nothing major, only a few comments according wiki ^^ @matteofigus

matteofigus:
Thanks @jankowiakmaria 
* Yes, desired
* Fixed
* Reduced the logo

Can you merge when you're happy?

Thanks

Issue #252, opened 2016-06-20
Multiple Security Issues


duereg:
All of these security issues are problems in OC:

"https://nodesecurity.io/advisories/39",
"https://nodesecurity.io/advisories/48",
"https://nodesecurity.io/advisories/57",
"https://nodesecurity.io/advisories/61"

Please update your dependencies to remove these issues.

PR #265, opened 2016-07-05
[careful + breaking change] Removed Handlebars 3 support

matteofigus:
This is the last PR for fixing #252 - following #243 

* This will trigger a new minor version as it's a breaking change
* All the handlebars 3 stuff has been removed


matteofigus:
I think this could be kind of ready if you want to start doing a review @antwhite / @mattiaerre?

src/components/oc-client/_package/src/oc-client.js
```diff
@@ -302,41 +301,16 @@ var oc = oc || {};
...
-
-                  callback(null, html);
-                });
               });
-
-              // End of the Handlebars "migrate" part
-
             } else if(compiledViewInfo.type === 'jade'){
-              oc.require('jade', JADE_URL, function(jade){
+              oc.require('jade', JADE_URL, function(){
```
mattiaerre: this is not related right? you just removed the parameter as not used by that function correct? a.k.a. "[the boy scout rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule)" 😄

src/components/oc-client/src/oc-client.js
```diff
@@ -302,41 +301,16 @@ var oc = oc || {};
...
-
-                  callback(null, html);
-                });
               });
-
-              // End of the Handlebars "migrate" part
-
             } else if(compiledViewInfo.type === 'jade'){
-              oc.require('jade', JADE_URL, function(jade){
+              oc.require('jade', JADE_URL, function(){
```
mattiaerre: same as above right?

matteofigus: Yes to both :)

client/src/validator.js
```diff
@@ -4,6 +4,20 @@ var settings = require('./settings');
 var _ = require('./utils/helpers');
 
 module.exports = {
+  validateComponent: function(template, options){
```
mattiaerre: what about adding a unit test around this behaviour?

mattiaerre:
not sure how it works w/ licenses but what about this?

https://github.com/opentable/oc/blob/handlebars4/src/components/oc-client/LICENSES#L80

test/front-end/render.js
```diff
@@ -37,24 +37,14 @@ describe('oc-client : render', function(){
       var originalHb3, originalHb4;
```
mattiaerre: I think we do not need these variables anymore do we?

matteofigus: Ok I cleaned up this

client/src/validator.js
matteofigus: Done 👍

antwhite:
LGTM

matteofigus:
Thanks! So, @mattiaerre can you merge when you're happy about the tests? (I added some as requested).

test/unit/client-validator.js
```diff
@@ -32,4 +32,43 @@ describe('client : validator', function(){
       });
     });
   });
+
+  describe('when validating component', function(){
```
mattiaerre: awesome 👍

mattiaerre:
good job @matteofigus 
thanks,

// cc @antwhite

PR #268, opened 2016-07-06
Strict shrinkwrapping

### LINKED ISSUE #267: shrinkwrap errors
Hi @matteofigus 

I seem to be getting an error when I try to shrinkwrap my project, which includes the oc-client.  It seems to  be failing as a result of the oc-client shrinkwrapping an invalid version of uglify-js.  Specifically, see

```javascript
  "uglify-js": {
      "version": "2.6.4",
      "from": "uglify-js@latest",
      "resolved": "https://registry.npmjs.org/uglify-js/-/uglify-js-2.6.4.tgz",
...
```

As you can see it requires the latest uglify-js, but resolves to 2.6.4, which is no longer the latest version.  We are still debugging, so we can't be certain this is our issue, but it seems likely.

cc: @maxgurewitz

matteofigus:
Fixes #267


jwhayes:
LGTM

matteofigus:
Note: merged despite AppVeyor is red, as they are dealing with some issues today.

PR #270, opened 2016-07-15
server-side nested rendering

### LINKED ISSUE #224: OC plugin to render nested components on the server-side
When rendering nested components on the server-side, an unnecessary network trip is required to make extra requests to the registry itself.
The ideal solution for optimally approaching to this scenario would be to have a plugin to render nested components using exposed registry internals to do the rendering via the same instance. 

The api for a `server.js` file should be something like this:
```js
module.exports.data = function(context, callback){
  var nestedComponents = [
    { name: 'component', parameters: { 'hello': 'world' }},
    { name: 'component2', parameters: { 'hello': 'world' }}
  ];
  context.plugins.renderComponents(nestedComponents, function(err, components){
    callback(null, components);
  });
};
```
The plugin should have the same API as [the node.js client's `renderComponents` method](https://github.com/opentable/oc/tree/master/client#clientrendercomponentscomponents--options-callback) - but it should do the rendering without making an extra request to itself.

matteofigus:
Fixes #224 

The plan is to expose in the `context` object 2 new methods, similar to the node.js client:

### context.renderComponent(componentName [, options], callback)
with options containing optional params `parameters`, `headers`, `version` and callback with a signature of (error, html)

### context.renderComponents(components [, options], callback)
with components array of objects containing mandatory `name` and optional `version` and `parameters`
and options containing optional params `headers` and callback with a signature of (error, html)


jankowiakmaria:
LGTM but I think I'd like to be added:
* more unit tests for rendering multiple components (currently it's always one in tests with `nestedRenderer.renderComponents`) with and without component specific and shared parameters,
* and acceptance test for `renderComponents` (currently there is one for `renderComponent`)

matteofigus:
Ok @jankowiakmaria I'll work on that - thanks!

jankowiakmaria:
LGTM, merging

PR #278, opened 2016-08-09
Container client override

matteofigus:
This allows to render multiple components with multiple container strategies.


antwhite:
LGTM

PR #280, opened 2016-08-14
Respect dependency versions for components

opoku:
Dependencies of components are always installed at the latest version even though a particular version might have been specified in their `package.json`.

This patch pulls in the version information from the `package.json` and forwards it to the `npm` so that the appropriate package version is installed.

Caveat: this patch needs to be made more generic so that all the formats supported by npm install are handled properly. I'm putting it up to start the discussion on the necessity of this work.


matteofigus:
Great, thanks for your help @opoku

Issue #284, opened 2016-09-06
npm install on dev watcher doesn't work when launching runner from container dir


matteofigus:
Probably introduced with #280 

It looks like npm install happens and succeeds, but then the api starts and seems not to be able to find the dependency and errors.

matteofigus:
Update: this is unrelated from #280

matteofigus:
This is now fixed.

PR #285, opened 2016-09-11
npm install and dep resolvement needs to happen in the same dir

matteofigus:
This PR fixes two issues:
* When working on #280 we introduced the possibility to npm install dependency@version instead of just the dependency on the latest version. The bug was that multiple components using the same dep where installed multiple times. In addition, resolving the dependency when running the component from registry resulted in a dependency bug because the dependency wasn't registered correctly. This should fix.
* The second is the fix for #284 - npm install succeeds but then goes to an infinite loop. This is actually not related to #280 and happens when doing `oc dev componentsFolder`. Basically, until now npm install was installing inside components folder, and dependency were resolved on the folder the `oc dev` was running. So, `oc dev .` was always working (so, when launching the dev inside the components folder) but `oc dev componentsFolder` wasn't. From now, with this fix, it is decided that npm dependencies are installed in a `node_modules` folder relative to the `oc dev` and that's where they are resolved. This should make the dev watcher able to run from both locations.

The easiest way to test this stuff is to use a folder of components using dependency and manually run the acceptance tests from various locations. We can also add extra unit testing but I am not too concerned as this is mostly a CLI minor feature and shouldn't have impact on a registry.

@lobut let's have a look at this together tomorrow so that we can do some manual testing together and then merge ;)

### DIFF
--- a/src/cli/domain/get-components-deps.js
+++ b/src/cli/domain/get-components-deps.js
@@ -7,10 +7,17 @@ var _ =  require('underscore');
 module.exports = function(components){
   var deps = [];
   _.forEach(components, function(c){
+
     var pkg = fs.readJsonSync(path.join(c, 'package.json'));
+
     _.forEach(_.keys(pkg.dependencies), function(d){
-      if(!_.contains(deps, d)){
-        deps.push(d + '@' + pkg.dependencies[d]);
+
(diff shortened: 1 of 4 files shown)


matteofigus:
Reviewed with @lobut so auto merging.

PR #286, opened 2016-09-11
Hot reload

### LINKED ISSUE #283: OC dev destroys the instance during hot reload and doesn't allow to test caching
It would be nice to have a parameter for the oc dev to preserve the created instance, in order to simulate production instance during testing (manual and automatic).

This is particularly useful for when an instance uses the memory for caching some small data.
For instance,

```js
var cache = null;
module.exports.data = function(cx, cb){
  if(cache) return cb(null, cache);

  // make an api call and save it to cache...
  cb(null, cache);

  // flush the cache after 5 mins
  setTimeout(function(){ cache = null; }, 5000 * 60);
};
```
As discussed with @nickbalestra and @lobut

matteofigus:
This fixes #283 

Basically allows to disable hot reloading (which is going to not look for changes and preserve the created instance simulating the live registry instance), via running 
`oc dev components/ 3030 --hotReloading=false`

Note: this depends on PR: #285

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,10 @@
 
 OpenComponents, **serverless in the front-end world**.
 
+OpenComponents is an open-source framework that allows fast-moving teams to easily build and deploy front-end components. It abstracts away complicated infrastructure and leaves developers with very simple, but powerful building blocks that handle scale transparently.
+
+#### How does it work?
+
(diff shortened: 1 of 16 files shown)


matteofigus:
Reviewed this with @lobut

PR #288, opened 2016-09-15
oc-client warmup improvements

matteofigus:
This PR is for improving the warmup mechanism in the oc node.js client
* Some unit tests to cover the basic functional aspects
* Fixed a little bug that in some scenarios could generate wrong urls to warmup 
* More detailed error log in case of failing request


test/unit/client-warmup.js
```diff
@@ -0,0 +1,152 @@
+'use strict';
+
+var expect = require('chai').expect;
+var injectr = require('injectr');
+var sinon = require('sinon');
+
+describe('client : warmup', function(){
```
mattiaerre: well done w/ the tests; they helped me to understand what the feature does

client/src/utils/helpers.js
```diff
@@ -17,12 +17,12 @@ var _ = {
   eachAsync: function(obj, fn, cb){
     var c = obj.length;
     
-    var next = function(){
+    var next = function(err){
```
mattiaerre: in this case, the variable names are not helping very much w/ readability aren't they?

matteofigus: True. This is really just a little custom polyfill for some little async functions that I use - mostly to keep the client as much "dependency free" and small as possible. The behaviour I changed here is that for a group of parallel async tasks, I want the cycle to exit not only when all the tasks are done - but **also** when any of the tasks errors via "calling back" an error.

test/unit/client-warmup.js
```diff
@@ -0,0 +1,152 @@
...
+        }
+      }, function(err){
+        error = err;
+        done();
+      });
+    });
+
+    it('should return an error with all the details', function(){
+
+      var expectedError = 'Error: Error warming up oc-client: request ' +
```
matteofigus: This is consistent with the other test. I don't mind tests being a bit verbose and dirty if it improves readability.

matteofigus: I cleaned up that to be a bit more readable btw hopefully it is better

client/src/utils/helpers.js
matteofigus: Renamed vars a bit to improve readability, hopefully this is a bit better

client/src/utils/helpers.js
```diff
@@ -15,14 +15,16 @@ var _ = {
...
     
-    var next = function(){
-      c--;
-      if(c === 0){ 
-        var a = cb;
+    var next = function(err){
+      callbacksLeft--;
+      if(callbacksLeft === 0 || !!err){
+
+        var cbCopy = cb;
```
mattiaerre: I do not quite understand why you do cbCopy (not entirely sure you do copy it btw) and also you do _.noop ... can you explain as I am really curious about it?

test/unit/client-warmup.js
```diff
@@ -0,0 +1,157 @@
...
+
+      var expectedRequest = {
+        url: 'https://my-registry.com/component1/~info',
+        json: true,
+        headers: { 'Accept-Language': 'en-US' },
+        method: 'GET',
+        timeout: 5
+      };
+
+      var expectedError = 'Error: Error warming up oc-client: request ' + JSON.stringify(expectedRequest) + ' failed (timeout)';
```
mattiaerre: I like that 👍

client/src/utils/helpers.js
matteofigus: This 2 lines are to ensure main callback is called just once. So basically I am overriding main callback with noop (function(){}) so that future executions do nothing. This is mostly for handling the scenario of when I fire the callback as soon as the error happens (so, other things could complete later but I ignore them).

mattiaerre: I understand, all good to me then; thanks!

matteofigus:
merging as I got the reviewers 👍

PR #291, opened 2016-09-19
Client errors

matteofigus:
This follows #288 - basically tackles 2 things:
1) When error happens for HTTP request to registry POST route, log contains all the request details
2) When HTTP request to registry happens and the requests times-out, it doesn't (erroneously) say it's a 500 anymore - but shows it is a timeout instead.


test/acceptance/client.js
```diff
@@ -341,13 +408,65 @@ describe('The node.js OC client', function(){
...
+        });
+
+        it('should contain a blank html response', function(){
+          expect(result).to.eql('');
+        });
+
+        it('should contain the error details', function(){
+
+          var exp = getRegExpFromJson(expectedRequest),
+              expected = new RegExp('Error: Server-side rendering failed: request ' + exp + ' failed \\(timeout\\)');
```
ArnoldZokas: @matteofigus is it worth including actual timeout value in the error message as well (for diagnostic purposes)

matteofigus: It is there. The client has a [default overridable timeout value](https://github.com/opentable/oc/blob/master/client/src/sanitiser.js#L44) that is always passed through the request call. Now we log all the arguments that we pass to that function call ( a JSON.stringify of `expectedRequest` - a couple of lines up) - which contains timeout, headers, body payload, everything.

An example of an error (I censored some private data stuff):

[Error: Server-side rendering failed: request {"url":"http://xxx","method":"post","headers":{"accept-language":"fr-CA, fr, en, *","ot-requestid":"xxx","ot-referringhost":"127.0.0.1","ot-referringservice":"xxx","accept":"application/vnd.oc.unrendered+json","user-agent":"oc-client-1.10.0/v4.4.4-darwin-x64"},"timeout":5,"json":true,"body":{"components":[{"name":"xxx","parameters":{xxx},"version":"1.2.0"}]}} failed (timeout)] ]

PR #300, opened 2016-10-05
[wip] Jade To Pug

matteofigus:
First part of #298.
More specifically, Step 1 will be to
- [ ] add Pug support to the registry
- [ ] add Pug support to the node.js client
- [ ] add Pug support to the browser client
- [ ] add Pug to the CLI
- [ ] remove Jade support from the CLI

This will allow the registry to be upgraded without breaking changes, clients to be upgraded, and component creators to migrate to Pug.

Step 2 will be to remove Jade support from the registry + remove Jade support from the client as final cleanup (separate PR).

Issue #297, opened 2016-09-30
Invalid dependencies


maxgurewitz:
The following,

```
npm i -g npm@3.10.8
cd oc/client
npm i
npm ls
```

outputs,

```
npm ERR! invalid: uglify-js@2.6.2 /Users/mgurewitz/develop/oc/client/node_modules/jade/node_modules/transformers/node_modules/uglify-js
npm ERR! extraneous: async@0.2.10 /Users/mgurewitz/develop/oc/client/node_modules/jade/node_modules/transformers/node_modules/uglify-js/node_modules/async
npm ERR! extraneous: source-map@0.5.6 /Users/mgurewitz/develop/oc/client/node_modules/jade/node_modules/transformers/node_modules/uglify-js/node_modules/source-map
npm ERR! extraneous: uglify-to-browserify@1.0.2 /Users/mgurewitz/develop/oc/client/node_modules/jade/node_modules/transformers/node_modules/uglify-js/node_modules/uglify-to-browserify
npm ERR! extraneous: yargs@3.10.0 /Users/mgurewitz/develop/oc/client/node_modules/jade/node_modules/transformers/node_modules/uglify-js/node_modules/yargs
```

These errors are causing shrinkwrap to fail in search-result-node, and preventing us from upgrading oc-client.  You can fix this by upgrading jade to the most recent version of pug.

matteofigus:
This is now fixed in `oc-client@1.11.1`

PR #301, opened 2016-10-05
Oc client jadeless

matteofigus:
This is a short term fix for #297 

Given oc-client requires Jade only for its own client-side runtime, and this is going to be removed soon anyway in favour of Pug on #300 (which compiles encapsulated templates so runtime is not gonna be needed anymore) - this should fix some problems as insecure dependencies and npm shrinkwrap. In fact, npm-shrinkwrap is not actually needed anymore in the oc-client as it was a temporary measure to get snyk green.


client/src/executor.js
```diff
@@ -4,7 +4,7 @@ var vm = require('vm');
 
 module.exports = {
   template: function(code, template){
-    var context = template.type === 'jade' ? { jade: require('jade/runtime.js')} : {};
+    var context = template.type === 'jade' ? { jade: require('./renderers/support/jade-runtime.js')} : {};
```
mattiaerre: so it looks like we are including the `jade-runtime` lib as a static resource right?

client/src/renderers/support/jade-runtime.js
```diff
@@ -0,0 +1,252 @@
+(function(f){if(typeof exports==="object"&&typeof module!=="undefined"){module.exports=f()}else if(typeof define==="function"&&define.amd){define([],f)}else{var g;if(typeof window!=="undefined"){g=window}else if(typeof global!=="undefined"){g=global}else if(typeof self!=="undefined"){g=self}else{g=this}g.jade = f()}})(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
```
mattiaerre: it this the file that you would find in `/node_modules/jade/lib/runtime.js`?

matteofigus: No, this is the one you would find in `/node_modules/jade/runtime.js`

client/src/executor.js
matteofigus: True, this is the only thing I need from Jade. Including it as whole dependency means including a series of sub-dependencies I don't care about which contains some known security vulnerabilities - which is the reason we shrinkwrapped the deps in the first place. This is going to be cleaned-up as part of the Pug migration.

client/src/renderers/support/jade-runtime.js
mattiaerre: kk gotcha!

mattiaerre:
LGTM 👍

Issue #272, opened 2016-07-20
Additional data when on rendering failure for POST endpoint


jankowiakmaria:
Currently when something goes wrong during rendering one of the components (when using the POST (batch) endpoint) the response looks like:
```
[
  {
    "status": 400,
    "response": {
      "code": "NOT_VALID_REQUEST",
      "error": "Expected mandatory parameters are missing: param"
    }
  },
  {
    "status": 200,
    "response": {
      "type": "oc-component",
      "version": "1.1.4",
      "requestVersion": "",
      "name": "successfully-rendered-component-name",
      ...
    }
  }
]
```
I think it would be useful to add `name` (and optionally `requestVersion`) to the response of component which failed to render. It would produce something like:
```
[
  {
    "status": 400,
    "response": {
      "code": "NOT_VALID_REQUEST",
      "error": "Expected mandatory parameters are missing: param",
      "name": "failed-to-render-component-name",
      "requestVersion": "1.2.X"
    }
  },
  {
    "status": 200,
    "response": {
      "type": "oc-component",
      "version": "1.1.4",
      "requestVersion": "",
      "name": "successfully-rendered-component-name",
      ...
    }
  }
]
```

PR #302, opened 2016-10-09
Component's name and request version in response when failure

jankowiakmaria:
As per #272
(It actually adds name and request version not only for POST endpoint)


matteofigus:
Excellent, thank you @jankowiakmaria

PR #303, opened 2016-10-10
Client#renderComponents accepts global parameters

### LINKED ISSUE #185: Client#renderComponents should accept global parameters
Currently when you render multiple components you should pass parameters for each of it separately. For performance reasons it would be better to pass one "parameters" object which can be used by all the components, e.g.:

```js
client.renderComponents([
    { name: 'header' },
    { name: 'footer', version: '4.5.X' },
    { name: 'advert', render: 'client' }
], {
  container: false,
  parameters: { 
    loggedIn: true
  },
  headers: {
    'accept-language': 'en-US'
  },
  timeout: 3.0
}, function(errors, htmls){ 
    // do stuff
});
```

jankowiakmaria:
fixes #185


test/acceptance/client.js
```diff
@@ -246,7 +351,8 @@ describe('The node.js OC client', function(){
           components: [{
             name: 'hello-world',
             version: '~1.0.0'
-          }]
+          }],
+          parameters: {}
```
matteofigus: uhm, why an empty `parameters` here? Just to test you can pass an empty object and that still works?

test/acceptance/client.js
```diff
@@ -367,7 +473,10 @@ describe('The node.js OC client', function(){
                 parameters: {
                   hi: 'john'
                 }
-              }]
+              }],
+              parameters: {
+                hi: 'john'
+              }
```
matteofigus: I guess same reason here (2)

test/acceptance/client.js
```diff
@@ -438,7 +547,11 @@ describe('The node.js OC client', function(){
                 errorType: 'timeout',
                 timeout: 1000
               }
-            }]
+            }],
+            parameters: {
+              errorType: 'timeout',
+              timeout: 1000
+            }
```
matteofigus: and here (3) ?

test/acceptance/client.js
jankowiakmaria: it's because of the way `renderComponent` works (https://github.com/opentable/oc/blob/master/client/src/index.js#L26-L43). It uses `renderComponents` behind the scenes (which by default passes empty object as global parameters) - that's why `expectedRequest` contains `parameters`.

In case of the other two tests - the same thing, we're using `renderComponents` and in options (for `renderComponent`) we pass parameters which means they become global in internal `renderComponents` call.

matteofigus:
Thanks, excellent!

PR #307, opened 2016-10-14
Removing spaces from decodedUrl params before passing for [GT-1372]

IRog:
https://opentable.atlassian.net/browse/GT-1372

Let me know, if there's any changes you want. i.e. maybe more sanitization or such


nchallapalli:
@matteofigus could you please review and let us know. Looks good to me.

matteofigus:
Whitespaces are legal, I know it's not nice, but I prefer to keep the API simple and delegate to clients to enforce more strict checks where appropriate.

PR #309, opened 2016-10-15
Default value for optional parameters

### LINKED ISSUE #276: Default value for optional params
I found my self doing this quite often for non mandatory params:

`var something = context.params.something || "defaultValue";`

I think it would be valuable being able to specify  (in the case of !mandatory parameters) a default value in the package.json under oc.parameters.

jankowiakmaria:
resolves #276

Would require change in [wiki -> Getting started](https://github.com/opentable/oc/wiki/Getting-started#structure-of-the-package) page:
|`oc.parameters[key].default`|`*`|Default value of optional parameter - applied when value is not specified in the request|


src/registry/routes/helpers/apply-default-values.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+var _ = require('underscore');
+
+module.exports = function(requestParameters, expectedParameters) {
+  var result = _.clone(requestParameters);
```
matteofigus: is this `clone` necessary? If a component has a decent number of params (like TMS) this clone could use some CPU. If not necessary, I would be tempted to remove it? What do you think

src/registry/routes/helpers/apply-default-values.js
```diff
@@ -0,0 +1,18 @@
...
+  var optionalParametersWithDefaults = _.pick(expectedParameters, function(parameter){
+    return !parameter.mandatory && parameter.default;
```
matteofigus: what when parameter.default is a boolean false? maybe a `!_.isUndefined(parameter.default)` would be better?

matteofigus:
🎉 🎂

matteofigus:
Thanks! I updated the wiki too.

jankowiakmaria:
👍

PR #313, opened 2016-10-21
Recursively look for oc.json, starting from componentsDir 

### LINKED ISSUE #219: oc dev looks for local oc.json but could do better than that
Basically,  when 'oc-components' is a folder of components, 
* `cd oc-components && oc dev . 3030` works (if oc.json is there)
* just `oc dev oc-component 3030` doesn't (because it looks for oc.json in that folder).

The cli should try to look for oc.json inside the folder and then in the parent if not found.

matteofigus:
Fixes #219


src/cli/domain/get-mocked-plugins.js
```diff
@@ -25,11 +25,12 @@ var registerStaticMocks = function(mocks, logger){
   });
 };
 
-var registerDynamicMocks = function(mocks, logger){
+var registerDynamicMocks = function(mocks, ocJsonPath, logger){
```
mattiaerre: I think you should rename this parameter into `ocJsonLocation` as per the way you invoke the method below in line 97

src/cli/domain/get-mocked-plugins.js
```diff
@@ -53,23 +54,47 @@ var registerDynamicMocks = function(mocks, logger){
...
+
+      return findPath(parentDir, fileName);
+    }
+  }
+  
+  return fileToResolve;
+};
+
+module.exports = function(logger, componentsDir){
+  componentsDir = path.resolve(componentsDir || '.');
```
mattiaerre: so if you do not pass the `componentsDir` parameter you set it to '`.` correct?

test/unit/cli-domain-get-mocked-plugins.js
```diff
@@ -2,45 +2,123 @@
...
+        expect(result.length).to.equal(3);
+      });
+    });
+
     describe('when oc.json is missing', function(){
       var result;
       beforeEach(function(){
         initialise({ existsSync: sinon.stub().returns(false) });
-        result = getMockedPlugins(console);
+        result = getMockedPlugins(console, '/root/components/');
```
mattiaerre: I see that now all the tests are passing the path parameter to the `getMockedPlugins` method; I think we should have tests w/o the second parameter as well what do you think?

src/cli/domain/get-mocked-plugins.js
```diff
@@ -53,23 +54,47 @@ var registerDynamicMocks = function(mocks, logger){
...
+
   if(!content.mocks || !content.mocks.plugins){
     return plugins;
   }
 
   logger.log(colors.yellow(strings.messages.cli.REGISTERING_MOCKED_PLUGINS));
 
   plugins = plugins.concat(registerStaticMocks(content.mocks.plugins.static, logger));
-  plugins = plugins.concat(registerDynamicMocks(content.mocks.plugins.dynamic, logger));
+  plugins = plugins.concat(registerDynamicMocks(content.mocks.plugins.dynamic, ocJsonLocation, logger));
```
mattiaerre: I am also thinking about preserve the same "pattern" between the methods (registerStaticMocks and registerDynamicMocks) such as adding the extra parameter either at the very beginning or after "logger" but not in between 😅 

`registerDynamicMocks(ocJsonLocation, content.mocks.plugins.dynamic, logger)`

or

`registerDynamicMocks(content.mocks.plugins.dynamic, logger, ocJsonLocation)`

src/cli/domain/get-mocked-plugins.js
matteofigus: correct.

test/unit/cli-domain-get-mocked-plugins.js
matteofigus: I can do it.

src/cli/domain/get-mocked-plugins.js
matteofigus: ok

mattiaerre:
LGTM 👍  (squashmerge)

PR #314, opened 2016-10-24
Revert async upgrade + timeout wrapping

matteofigus:
This reverts #310 as new version of async seems not very performant compared to previous and wrapping calls seems not to be useful.


ArnoldZokas:
@matteofigus LGTM. We could try other, older 2.x.x releases?

matteofigus:
@ArnoldZokas definitely want to do, plan was to rollback this one first as fix and then upgrade in the future as I know previous was more stable 👍

PR #317, opened 2016-10-31
Client dynamic hrefs

matteofigus:
Allows to override registry baseUrls on each `renderingComponent` or `renderingComponents` call.


ArnoldZokas:
LGTM

Issue #322, opened 2016-11-20
Unable to require nested paths


matthewdavidson:
The following code in `server.js` does not work, even though `react-dom` is listed as a dependency in 
the components `package.json`:
```js
var ReactDOMServer = require('react-dom/server');
```
I get the following error:
```sh
Packaging components...an error happened while packaging <component_path>: 
Missing dependencies from package.json => ["react-dom/server"]
```

Here is a reproduction of the issue: https://github.com/matthewdavidson/oc-nested-requires-bug

I'm happy to attempt a PR unless you have any major objections :).

matteofigus:
Hi @matthewdavidson thanks for submitting the issue, good catch!

Definitely I bug I would like to be fixed! I would love a PR for that 👍

PR #323, opened 2016-11-22
Fix for #322

matthewdavidson:
This PR is by no means complete - I've spent a bit of time familiarising myself with the codebase and this is my quick hack to get things working. However as a result I have some observations...

Whilst I guess it's relatively uncommon, lodash is a good example of a package that exposes entry points multiple levels deep such as `lodash/fp/object` or `lodash/fp/curryN`. The method currently used to resolve packages - by reading the components `package.json` - wont be enough to future proof this use case as it's not feasible to statically analyse the package for all of the possible entry points (at n levels of nesting).

An alternative approach would be to refactor the `require-wrapper` to pass any paths through to `require` as long as they match dependencies that are listed in the components `package.json` as well as conforming to the registries whitelist of dependencies.

@matteofigus thoughts?


matteofigus:
Hi @matthewdavidson thanks for the PR. I agree about your proposal.

Summarising, I would be happy if both component and registry maintainer declare "lodash" as dependency, and then server.js is able to require anything that matches "lodash" or "lodash/anything" or "lodash/anything/anything(etc)...".

matthewdavidson:
Apologies for the delay - I've updated the PR. 

`dependency-resolver.js` has been ditched in favor of `require-wrapper.js` serving only to validate any requires against the dependency whitelist.

npm-shrinkwrap.json
```diff
@@ -3066,6 +2989,11 @@
         }
       }
     },
+    "require-package-name": {
+      "version": "2.0.1",
+      "from": "require-package-name@latest",
+      "resolved": "https://registry.npmjs.org/require-package-name/-/require-package-name-2.0.1.tgz"
+    },
```
matthewdavidson: Done (i think).

matteofigus: Thanks, can you just pull in the latest master so we can get that extra test ;)

matteofigus:
Thanks for taking care of this @matthewdavidson - just published this to a new version

matthewdavidson:
👍 🎉

Issue #325, opened 2016-11-28
Add optional cache headers for components


ivan-dimitrov-skyscanner:
We are thinking of adding some optional cache headers per component. For instance if you ask for http://..registry../hello-world/1.0.0 you would get some cache-control headers and benefit from CDNs or browser caches. We could use the component's package.json file as a carrier of those additional headers.

For example for hello-world component the package.json file could look like:

```javascript
{
  "name": "hello-world",
  "description": "",
  "version": "1.0.0",
  "repository": "",
  "oc": {
    "files": {
      "data": "server.js",
      "template": {
        "src": "template.html",
        "type": "handlebars"
      }
    }
  },
  "headers": {
    "Cache-Control": "public max-age=3600",
    "Custom-Header": "custom-header-value"
  }
}
```

Then this **headers** section could be merged with the actual HTTP response headers but could also be added in the component's metadata:

(1) When rendered:
```javascript
{
  "type": "oc-component",
  "version": "1.0.0",
  "requestVersion": "1.0.0",
  "name": "hello-world",
  "renderMode": "rendered",
  "headers": {
    "Cache-Control": "public max-age=3600",
    "Custom-Header": "custom-header-value"
  },
  "href": "http://registry_url/hello-world/1.0.0",
  "html": "...component-html..."
}
```

and (2) When unrendered:
```javascript
{
  "type": "oc-component",
  "version": "1.0.0",
  "requestVersion": "1.0.0",
  "name": "hello-world",
  "renderMode": "unrendered",
  "headers": {
    "Cache-Control": "public max-age=3600",
    "Custom-Header": "custom-header-value"
  },
  "href": "http://registry_url/hello-world/1.0.0",
  "data": {
    
  },
  "template": {
    "src": "http://registry_url/hello-world/1.0.0/static/template.js",
    "type": "handlebars",
    "key": "737b5c8ac8536a42972facd7f2bbfc945326df42"
  }
}
```

However this is going to work for single components only. In case of bulk requests we could still return the headers entry in the components' metadata for use in the client if necessary but there is no point of setting cache headers in the HTTP response (as the request is POST and it normally will not be cached by the intermediate proxies).

What are your thoughts on that?

matteofigus:
Hi @ivan-dimitrov-skyscanner I would like for components to have the ability of setting response headers. I was talking about this feature with @seif and @antwhite too.

[...]

PR #326, opened 2016-12-02
Add optional cache headers for components (#325)

ivan-dimitrov-skyscanner:
* Added a function setHeader(name, value) to the context object that we pass to the component data function.

* Subsequently these headers (if added) will be set in the response data for the returned component JSON as well as normal HTTP response headers (component.js)

* Added a mock component (response-headers-component) and a test to cover the new functionality


matteofigus:
Awesome stuff @ivan-dimitrov-skyscanner, thanks for the PR!

Issue #327, opened 2016-12-08
Custom cache headers set by one component could appear as a cache headers for another


ivan-dimitrov-skyscanner:
@matteofigus 

We found a bug introduced in #326 - the problem is on line 30 in get-component.js file. There we introduced responseHeaders variable but this is in a outerer context than it should be. Because of this every invocation of render sees one and the same variable and every component inherits the headers introduced by former ones.
The fix is to move the responseHeaders variable inside renderer function.

I'm basically having the fix and am working on a regression test to validate it. Perhaps tomorrow I'll do a PR.

matteofigus:
Ok, thanks. We haven't tested this version yet but planning to do next week. In case I'll wait for the fix!

Thanks

PR #328, opened 2016-12-09
(327) Fix - Custom headers of one component can appear to another.

ivan-dimitrov-skyscanner:
The problem was introduced with #326 -
In get-component.js the declaration of responseHeaders variable happens in a more
global scope than it should be. For this reason every invocation of renderer
function sees the changes in the headers made by previously requested
components.
The fix is to just move the declaration one scope inner.

- Added a unit test to validate the fix - it tries to load two components
subsequently. The first one does provide custom headers while the second
one - don't. Then we check and expect to see there are no custom headers
in the result for the second one.


matteofigus:
Excellent, thanks @ivan-dimitrov-skyscanner

matteofigus:
This fixed #327

Issue #324, opened 2016-11-23
Add endpoint for getting list of components with their versions


ferewuz:
Following up on discussion in gitter channel, to enable clients to always make versioned requests to oc-registry it would be useful having a separate endpoint returning all components with their versions included.

Implementation would be similar to `list-components` when called without `text\html` headers (https://github.com/opentable/oc/blob/master/src/registry/routes/list-components.js)

My thinking was adding a route called `list-components-versions`.
Return value will be a json, containing components and sorted list of versions, something like:
```
"components": {
    "http://oc-registry.com/component1" : [ '1.0.0', '1.1.0', '2.0.0'],
    "http://oc-registry.com/component2" : [ '1.0.0', '2.1.0', '3.0.0']
}
```

Possible extension would be to limit the components you want to get the version for, using query params. Not a must-have for first version.

I am happy to go forward with implementation and PR if that makes sense to you @matteofigus and other contributors?

matteofigus:
Hi @ferewuz, thanks for the issue. I spent some time to think on this, and I actually may have another idea.

The goal of this feature is to enable clients to understand what matches a requested version and understand if something changed given an amount of time. In theory, you can already do this, but I would like your opinion on that.

This is an example for the batch endpoint request, already implemented:

```js
POST https://oc-registry.com
{
  components: [{
    name: "component1",
    version: "1.x.x"
  }, {
    name: "component2",
    version: "1.4.x",
    parameters: { a: 12354 }
  }]
}

//would give you back something like:

{
  components: [{
    status: 200,
    response: {
      name: "component1",
      href: "https://oc-registry.com/component1",
      html: "...",
      requestVersion: "1.x.x",
      version: "1.2.3"
    }
  }, {
    status: 200,
    response: {
      name: "component2",
      href: "https://oc-registry.com/component2?a=12345
      html: "...",
      requestVersion: "1.4.x",
      version: "1.4.3"
    }
  }]
}
```

The good things about using the existing batch endpoint to achieve the same result:
1) You can ask only for the components you care about
2) You delegate to the registry the version resolving (with the other approach you may need to pull the `semver` module into the clients to handle to currently supported semver scenarios and perform the matching on the client side) - so clients can just compare if resolved versions are the same compared to a previous request

Bad things:
1) You get more than you want, like the component response, etc.
2) If a component has mandatory parameters etc, you actually need to include in the payload those parameters for of a proper request

So my idea is, what about modifying the existing BATCH endpoint to do what we want here.
[...]

PR #332, opened 2016-12-20
Add endpoint for getting list of components with their versions (#324)

ivan-dimitrov-skyscanner:
1. Changed the registry to anticipate 'Accept: application/vnd.oc.info+json'
HTTP header and if detect it to skip the component rendering and to return
just the component info.

2. Added a new method in Client - getComponentsInfo.
It prepares the request to the registry with the proper accept header, checks
for errors and provides a processed pair (error, info).
- in case all the components are valid and there is info data for all of them then the error
will be null and the info will have the following structure:
[{
   componentName: '<name>',
   requestedVersion: '<requested version|undefined>'
   apiResponse: {  //// comes from the registry
      href: '<component url>',
      name: '<name>',
      renderMode: '<rendered|unrendered>',
      requestVersion: <requested version>,
      type: '<type>',
      version: '<actual version>'
   }
},
{
   componentName: '<name>',
   requestedVersion: '<requested version|undefined>'
   apiResponse: {...}
},
...
]

- in case the request could not be satisfied for a certain component then its info part will not contain apiResponse and will include an error field:
{
   componentName: '<name>',
   requestedVersion: '<requested version|undefined>'
   error: ...
}

In addition to this the 'error' parameter in the callback will be an array with
the error elements corresponding to each component requested. If there's no
error for a particular component then the corresponding element in the 'error'
array will be null:
[
   null,  //// means the first component is OK
   { message: ..., stackTrace: ... },
   null,  //// third components is OK as well
   ...
]

### DIFF
--- a/client/src/get-components-info.js
+++ b/client/src/get-components-info.js
@@ -1,0 +1,99 @@
+'use strict';
+
+var format = require('stringformat');
+var request = require('minimal-request');
+
+var settings = require('./settings');
+var sanitiser = require('./sanitiser');
+var _ = require('./utils/helpers');
+
+module.exports = function(config) {
+
+  return function(components, options, callback) {
+    if (!_.isArray(components)) {
+      components = [components];
+    }
+
+    if (_.isFunction(options)) {
+      callback = options;
+    }
+
+    options = sanitiser.sanitiseGlobalGetInfoOptions(options);
+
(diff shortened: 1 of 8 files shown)


ivan-dimitrov-skyscanner:
Found a way to accomplish it with a little-less invasive change :-)
Will re-issue it soon

PR #335, opened 2016-12-21
Add endpoint for getting list of components with their versions (#324)

ivan-dimitrov-skyscanner:
1. Changed the registry to anticipate 'Accept: application/vnd.oc.info+json'
HTTP header and if detects it to skip the component rendering and to return
the component's info only.

2. Added a new method in the Client - getComponentsInfo(...)
It prepares the request to the registry with proper accept headers, checks
for errors and provides a processed result pair (error[], info[]).
- in case info data is provided for all the requested components then the error
will be null and the info will have the following structure:
[{
   componentName: '<name>',
   requestedVersion: '<requested version|undefined>'
   apiResponse: {  //// comes from the registry
      href: '<component url>',
      name: '<name>',
      renderMode: '<rendered|unrendered>',
      requestVersion: <requested version>,
      type: '<type>',
      version: '<actual version>'
   }
},
{
   componentName: '<name>',
   requestedVersion: '<requested version|undefined>'
   apiResponse: {...}
},
...
]

- in case of errors then the info part will skip the apiResponse and will
include an error field:
{
   componentName: '<name>',
   requestedVersion: '<requested version|undefined>'
   error: ...
}

In addition to this the 'error' will be an array holding the
errors corresponding to each component requested. If there's no
error for a particular component then the corresponding element in the 'error'
array will be null:
[
   null,  //// means the first component is OK
   { message: ..., stackTrace: ... },
   null,  //// third components is OK as well
   ...
]

### DIFF
--- a/src/registry/routes/component.js
+++ b/src/registry/routes/component.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var GetComponentHelper = require('./helpers/get-component');
-var Settings = require('../../resources/settings');
 
 module.exports = function(conf, repository){
 
--- a/src/registry/routes/component.js
+++ b/src/registry/routes/component.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var GetComponentHelper = require('./helpers/get-component');
+var Settings = require('../../resources/settings');
 
 module.exports = function(conf, repository){
 
--- a/src/registry/routes/helpers/get-component.js
+++ b/src/registry/routes/helpers/get-component.js
@@ -129,6 +129,7 @@ module.exports = function(conf, repository){
         }, conf.baseUrl);
 
         var isUnrendered = options.headers.accept === settings.registry.acceptUnrenderedHeader,
+            skipRendering = options.headers.accept === settings.registry.acceptInfoHeader,
             renderMode = isUnrendered ? 'unrendered' : 'rendered';
 
         var response = {
@@ -149,7 +150,12 @@ module.exports = function(conf, repository){
           renderMode: renderMode
         });
 
-        if(isUnrendered){
+        if (skipRendering) {
+          callback({
+            status: 200,
+            response: response
+          });
+        } else if (isUnrendered) {
           callback({
             status: 200,
             response: _.extend(response, {
--- a/src/registry/routes/helpers/get-component.js
+++ b/src/registry/routes/helpers/get-component.js
@@ -76,6 +76,19 @@ module.exports = function(conf, repository){
         });
       }
 
+      // Skip rendering and return only the component info in case of 'accept: application/vnd.oc.info+json'
+      if (options.headers.accept === settings.registry.acceptInfoHeader) {
+        return callback({
+          status: 200,
+          response: {
+            type: conf.local ? 'oc-component-local' : 'oc-component',
+            version: component.version,
+            requestVersion: requestedComponent.version,
+            name: requestedComponent.name,
+          }
+        });
+      }
+
       // check component requirements are satisfied by registry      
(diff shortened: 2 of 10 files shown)


src/registry/routes/component.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 var GetComponentHelper = require('./helpers/get-component');
+var Settings = require('../../resources/settings');
```
matteofigus: Looks like this line is not actually needed?

src/registry/routes/helpers/get-component.js
```diff
@@ -149,7 +150,12 @@ module.exports = function(conf, repository){
           renderMode: renderMode
         });
 
-        if(isUnrendered){
+        if (skipRendering) {
+          callback({
+            status: 200,
+            response: response
```
matteofigus: This includes `renderMode` which is not actually very relevant here. Would you mind omitting that property in case `skipRendering===true`?

matteofigus: Also, the problem of doing this here (instead of around line 78) is that here we go through validation and data closure execution, which I believe is not needed for the info purposes. I think we don't want to just skip rendering, but skip execution at all.

test/unit/registry-routes-components.js
```diff
@@ -104,6 +115,81 @@ describe('registry : routes : components', function(){
...
+    });
+
+    it('should return a response containing components in the correct order', function(){
+      expect(response[0].response.href).to.be.undefined;
+      expect(response[1].response.href).to.be.equal('http://components.com/async-error2-component/1.0.0');
+    });
+
+    it('should return a response with error code and description for the first component', function(){
+      expect(response[0].response.code).to.be.equal('GENERIC_ERROR');
+      expect(response[0].response.error).to.be.equal('Component execution error: thisDoesnotExist is not defined');
```
matteofigus: As previous comment says, I actually am not sure we want this here. I think info should just skip execution at all and give us just the versions. What do you think?

src/registry/routes/helpers/get-component.js
ivan-dimitrov-skyscanner: Agree with you -- will move it around line 78 to skip most of the stuff done further in the file.

src/registry/routes/component.js
ivan-dimitrov-skyscanner: Fixed

src/registry/routes/helpers/get-component.js
ivan-dimitrov-skyscanner: Done.

test/unit/registry-routes-components.js
```diff
@@ -104,6 +115,97 @@ describe('registry : routes : components', function(){
...
+    // it('should return name and request version for both components', function(){
+    //   expect(response[0].response.name).to.be.equal('async-error2-component');
+    //   expect(response[0].response.requestVersion).to.be.equal('1.X.X');
+    //   expect(response[1].response.name).to.be.equal('async-error2-component');
+    //   expect(response[1].response.requestVersion).to.be.equal('1.0.0');
+    // });
+
+    // it('should rerurn the actual version for the second component', function() {
+    //   expect(response[1].response.version).to.be.equal('1.0.0');
+    // });
```
matteofigus: Can we get rid of those comments? I think now it is correctly resolving the versions which is what I would expect from info endpoints so those tests are not necessary anymore

ivan-dimitrov-skyscanner: Ops, sorry about them. They weren't meant to be there.
Fixed!

matteofigus:
Amazing, thanks!

Issue #336, opened 2016-12-23
Make S3 Key/Secret optional to allow support  for "IAM role based access to s3 from EC2"


BhautikDoshi:
IAM role based access to s3 from EC2 (instead of storing s3 key/secret)
http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html

BTW, thanks for oc - trying it out for a big UI project mashing new/old UI coming from different projects/groups with version matrix. OC seems to be a great fit.

So to fix the s3 key/secret validation error during registry startup, I just had to remove the validation code by changing line 66 of oc/src/registry/domain/validators/registry-configuration.js
from:

    if(!conf.s3 || !conf.s3.bucket || !conf.s3.key || !conf.s3.region || !conf.s3.secret){

to:

    if(!conf.s3 || !conf.s3.bucket || !conf.s3.region ){

After the change, the oc-registry on EC2 comes up fine with just bucket name, region.
(The Javascript AWS SDK will pull the Key/Secret from EC2 instance using  http://169.254.169.254/latest/metadata/iam/security-credentials/role-name  - these key/secret are automatically rotated and SDK will refresh it - more secure than having to handle s3 credentials ourselves)

  s3: {
    bucket: '<BUCKET_NAME>',
    region: 'us-east-1',
    componentsDir: 'components'
  },

Still new to the OC code and not very familiar yet. But if this above code change seems fine, let me know if you like me to post a PR. Thanks.

matteofigus:
Hi @BhautikDoshi this is great, I actually wasn't aware of this.

If you could make a PR to make those properties optional as you recommend it would be awesome!

Also, thanks for your nice words about our work. Very glad to hear it is useful to you 👍

BhautikDoshi:
Hi Matteo, I have pushed the PR. Ran tests successfully. 
Look forward to contributing more as and when need arises to enhance OC for our use case which could benefit others.

matteofigus:
Fixed with #337

PR #337, opened 2016-12-29
Fix for #336 - Make S3 Key/Secret optional to allow support for "IAM …

BhautikDoshi:
#336 

Matteo, Here is the PR. I made Key/Secret pair optional. But also made sure that if either one is specified then other one is specified too. Added Test case as well.


matteofigus:
I just published this to a new version and updated the docs: https://github.com/opentable/oc/wiki/Registry#registry-configuration

Thanks for your help!

BhautikDoshi:
Thanks Matteo for pushing out new revision. Will update to 0.33.29.

PR #343, opened 2017-01-06
333 - Implement customHeadersToSkipOnWeakVersion functionality

ivan-dimitrov-skyscanner:
- Introduced a new registry option - customHeadersToSkipOnWeakVersion, which expects
to be an array of strings denoting the headers to be omitted from
the response in case a weak version is requested (i.e 1.x.x). The rational
behind this is the need to control cache related headers - in case 1.x.x
version is requested we would not like cache headers to be returned because
otherwise the response will be cached. Next time when someone asks for 1.x.x
it will get the cached response even though a new version may be available.
- by default customHeadersToSkipOnWeakVersion  is set to be empty.
- added unit and integration tests.

### DIFF
--- a/src/registry/routes/component.js
+++ b/src/registry/routes/component.js
@@ -7,20 +7,6 @@ module.exports = function(conf, repository){
 
   var getComponent = new GetComponentHelper(conf, repository);
 
-  var setCustomHeaders = function(req, res, componentResponse) {
-    var headersToSet = componentResponse.headers;
-
-    if (req.params.componentVersion !== componentResponse.version && 
-        !_.isEmpty(res.conf.customHeadersToSkipOnWeakVersion)) 
-    {
-      headersToSet = _.omit(headersToSet, res.conf.customHeadersToSkipOnWeakVersion);
-    }
-    
-    if (!_.isEmpty(headersToSet)) {
-      res.set(headersToSet);
-    }
-  };
-
   return function(req, res){
     getComponent({
       conf: res.conf,
@@ -34,8 +20,12 @@ module.exports = function(conf, repository){
         res.errorDetails = result.response.error;
       }
 
-      if (result.response.headers) {
-        setCustomHeaders(req, res, result.response);
+      if (!_.isEmpty(result.response.headers)) {
+        res.set(result.response.headers);
+        
+        if (req.method === 'GET') {
+          delete result.response.headers;
+        }
       }
 
       return res.json(result.status, result.response);
--- a/src/registry/routes/component.js
+++ b/src/registry/routes/component.js
@@ -1,10 +1,31 @@
 'use strict';
 
 var GetComponentHelper = require('./helpers/get-component');
+var _ = require('underscore');
 
 module.exports = function(conf, repository){
 
   var getComponent = new GetComponentHelper(conf, repository);
+
+  var setCustomHeaders = function(req, res, componentResponse) {
+    if (req.params.componentVersion === componentResponse.version) {
+      //strong version request
+      res.set(componentResponse.headers);
+    } else {
+      //weak version request, therefore skip the blacklisted headers if any
+      if (_.isEmpty(res.conf.customHeadersToSkipOnWeakVersion || [])) {
+          res.set(componentResponse.headers);
+      } else {
+        var headersToSkip = new Set(res.conf.customHeadersToSkipOnWeakVersion);
+
+        _.forEach(Object.keys(componentResponse.headers), function(header) {
+          if (!headersToSkip.has(header.toLowerCase())) {
+            res.set(header, componentResponse.headers[header]);
+          }
+        });
+      }
+    }
+  };
 
   return function(req, res){
     getComponent({
@@ -20,7 +41,7 @@ module.exports = function(conf, repository){
       }
 
       if (result.response.headers) {
-        res.set(result.response.headers);
+        setCustomHeaders(req, res, result.response);
       }
 
       return res.json(result.status, result.response);
--- a/src/registry/routes/component.js
+++ b/src/registry/routes/component.js
@@ -8,22 +8,16 @@ module.exports = function(conf, repository){
   var getComponent = new GetComponentHelper(conf, repository);
 
   var setCustomHeaders = function(req, res, componentResponse) {
-    if (req.params.componentVersion === componentResponse.version) {
-      //strong version request
-      res.set(componentResponse.headers);
-    } else {
-      //weak version request, therefore skip the blacklisted headers if any
-      if (_.isEmpty(res.conf.customHeadersToSkipOnWeakVersion || [])) {
-          res.set(componentResponse.headers);
-      } else {
-        var headersToSkip = new Set(res.conf.customHeadersToSkipOnWeakVersion);
+    var headersToSet = componentResponse.headers;
 
-        _.forEach(Object.keys(componentResponse.headers), function(header) {
-          if (!headersToSkip.has(header.toLowerCase())) {
-            res.set(header, componentResponse.headers[header]);
-          }
-        });
-      }
+    if (req.params.componentVersion !== componentResponse.version && 
+        !_.isEmpty(res.conf.customHeadersToSkipOnWeakVersion)) 
+    {
+      headersToSet = _.omit(headersToSet, res.conf.customHeadersToSkipOnWeakVersion);
+    }
+    
+    if (!_.isEmpty(headersToSet)) {
+      res.set(headersToSet);
     }
   };
 
--- a/test/acceptance/registry.js
+++ b/test/acceptance/registry.js
@@ -103,7 +103,7 @@ describe('registry', function(){
       it('should return the component with custom headers', function() {
         expect(result.version).to.equal('1.0.0');
         expect(result.name).to.equal('hello-world-custom-headers');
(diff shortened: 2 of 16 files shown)


src/registry/routes/component.js
```diff
@@ -1,11 +1,32 @@
...
 
   var getComponent = new GetComponentHelper(conf, repository);
 
+  var setCustomHeaders = function(req, res, componentResponse) {
+    if (req.params.componentVersion === componentResponse.version) {
+      //strong version request
+      res.set(componentResponse.headers);
+    } else {
+      //weak version request, therefore skip the blacklisted headers if any
+      if (_.isEmpty(res.conf.customHeadersToSkipOnWeakVersion || [])) {
```
matteofigus: Isn't  `|| []` redundant as you're already doing that in the options' sanitiser?

src/registry/routes/component.js
```diff
@@ -1,11 +1,32 @@
...
+          res.set(componentResponse.headers);
+      } else {
+        var headersToSkip = new Set(res.conf.customHeadersToSkipOnWeakVersion);
```
matteofigus: Wouldn't be easier to use [underscore's omit](http://underscorejs.org/#omit) here?
```
var headersToSet = _.omit(componentResponse.headers.map(function(el){
  return el.toLowerCase();
}), res.conf.customHeadersToSkipOnWeakVersion);
if(!_.isEmpty(headersToSet)) res.set(headersToSet);
```

test/acceptance/registry.js
```diff
@@ -46,6 +51,136 @@ describe('registry', function(){
...
 
+  describe('GET /hello-world-custom-headers', function() {
+
+    var httpGetRequest = function(options, callback) {
+      var callbackDone = false;
+      var requestData = url.parse(options.url);
+      requestData.headers = options.headers || {};
+      requestData.method = 'get';
+
+      var req = require('http').request(requestData).on('response', function(response) {
```
matteofigus: I guess you're doing all of this because `minimal-request` doesn't allow you to access response headers. This is not great, I can totally extend minimal-request to do that and then this test should look a bit more neat. Still, not a blocker for merging this, definitely a quick improvement for a later PR. Sorry for that 👍

src/registry/routes/component.js
```diff
@@ -1,11 +1,32 @@
...
+
+        _.forEach(Object.keys(componentResponse.headers), function(header) {
+          if (!headersToSkip.has(header.toLowerCase())) {
```
matteofigus: I was thinking that if we change [setHeader to sanitise the headers by "lowercasing" it](https://github.com/opentable/oc/blob/74792a8ecfa6c0cf04d0a4ce9d5eeecabe48a039/src/registry/routes/helpers/get-component.js#L249) then we may not need to "lowercase" it here?

src/registry/routes/component.js
ivan-dimitrov-skyscanner: Fixed

src/registry/routes/component.js
ivan-dimitrov-skyscanner: Fixed! Thanks for the idea ;-)

src/registry/routes/component.js
ivan-dimitrov-skyscanner: Done

test/acceptance/registry.js
ivan-dimitrov-skyscanner: Completely agree with you. Will modify/fix this once we have the change in minimal-request.

ivan-dimitrov-skyscanner:
@matteofigus, all the requested changes are in place now. This one also includes upgrading of minimal-request to version 2.2.0 :-)

test/unit/registry-routes-component.js
```diff
@@ -465,6 +465,53 @@ describe('registry : routes : component', function(){
     });
   });
 
+  describe('when getting a component with server.js that sets custom headers with empty customHeadersToSkipOnWeakVersion', function() {
```
matteofigus: "not empty"?

test/unit/registry-routes-component.js
```diff
@@ -465,6 +465,53 @@ describe('registry : routes : component', function(){
...
+    });
+
+    it('should return 200 status code', function() {
+      expect(code).to.be.equal(200);
+    });
+
+    it('should not set the HTTP response test-headers', function() {
+      expect(response.headers).to.not.be.null;      
+      expect(response.headers['test-header']).to.equal('test-value');
+      expect(headers).to.be.undefined;
```
matteofigus: I think here I got a bit confused by the response.headers and headers. Do we need here to test response.headers if what we care about is just `expect(headers).to.be.undefined` to prove that `should not set the HTTP response test-headers`?

Also, I think same would apply for line 456

npm-shrinkwrap.json
```diff
@@ -1414,8 +1414,8 @@
     },
     "minimal-request": {
       "version": "2.1.1",
-      "from": "minimal-request@2.1.1",
-      "resolved": "https://registry.npmjs.org/minimal-request/-/minimal-request-2.1.1.tgz"
+      "from": "minimal-request@2.2.0",
+      "resolved": "https://registry.npmjs.org/minimal-request/-/minimal-request-2.2.0.tgz"
```
matteofigus: I actually have done that on master already with #349 so I guess you just need to rebase from master here?

matteofigus: Oh I actually was able to rebase here. Hope you don't mind 👍

test/unit/registry-routes-component.js
ivan-dimitrov-skyscanner: Yea, I see your point.
The idea of skipping some predefined headers from the response is to cope with the normal HTTP proxies that stay on the request-response route. Because of the GET request if they are going to do something with the response (like caching) they will do it based on the request string where we don't have a strong version. That's why we validate we don't set the HTTP response headers.
On the other hand in the response body we do have both the requested version and the actually returned version. That's why I'm exposing the headers there as well, so the response is something like:
```js
{
  "type": "oc-component-local",
  "version": "1.0.1",
  "requestVersion": "1.0.1",
  "name": "hello-world",
  "renderMode": "rendered",
  "href": "http://localhost:3030/hello-world/1.0.1",
  "headers": {
    "test-header": "test-value"
  },
  "html": "...."
}
```

This allows us to have those custom headers in POST responses as well :-)

ivan-dimitrov-skyscanner: But more and more I'm thinking over this and more I agree with your note here...
Regarding the consistency perhaps it's better to strip them from everywhere.
If you agree I can issue one more commit with this in place?

matteofigus: I think I can't see any value for having `headers` into the GET /component response body, as we should use the headers for that. So, for simplicity, I agree about stripping the headers from the headers :rofl: in case this setting is on, and I vote for just removing the `headers` from the response in any case.

In case of the POST batch, I think it is a bit different. Using the response body for the headers is the simplest way to include the headers in each component. The way I would tackle it there, is to have a `headers` key in the response that behaves exactly as the headers in the GET. Still, I think we can do the POST in a separate PR if you agree.

ivan-dimitrov-skyscanner: I updated the change to include this -- I also put the logic for POST requests together with some tests

test/unit/registry-routes-component.js
ivan-dimitrov-skyscanner: Should be fixed

test/acceptance/registry.js
```diff
@@ -46,6 +51,228 @@ describe('registry', function(){
...
+
+        it('should not set HTTP custom headers', function() {
+          expect(headers).to.not.have.property('cache-control');
+          expect(headers).to.not.have.property('test-header');
+        });
+
+        it('should skip Cache-Control header', function() {
+          expect(result[0].response.version).to.equal('1.0.0');
+          expect(result[0].response.name).to.equal('hello-world-custom-headers');
+          expect(result[0].response.headers).to.be.deep.equal({'cache-control': 'public max-age=3600', 'test-header': 'Test-Value'});
```
matteofigus: Shouldn't this skip 'Cache-Control' header and be eql to `{'test-header': 'Test-Value'}}`?

matteofigus: Shouldn't this skip 'Cache-Control' header and be eql to `{'test-header': 'Test-Value'}`?

ivan-dimitrov-skyscanner: You are right. It's more on that though. Actually I didn't realise the POST requests go to component*s*.js instead of component.js. Now this is fixed. Also one of the big trouble with the tests for POST requests was that they are following after those for GET and they are working with an already set customHeadersToSkipOnWeakVersion. So I did a slight refactoring of the before/afters there as well.

Apologies for this mess :-)
And can I ask for one more review, please?

matteofigus: No problem, looking now 👍

matteofigus:
Amazing work @ivan-dimitrov-skyscanner - I published and updated the docs here: https://github.com/opentable/oc/wiki/Registry#registry-configuration - what do you think about the description?

ivan-dimitrov-skyscanner:
Awesome :-)
Thank you very much @matteofigus.
I like the description in the docs. Perhaps I have to put some example to document setHeader(...) method in the context.

matteofigus:
Thanks to you! Feel free to edit the wiki if you want, you should be enabled to do it ;)

Issue #340, opened 2017-01-03
Consider support for local requires in server.js


matthewdavidson:
Just ran into this ... 
```sh
Packaging components...an error happened while packaging /path/to/my/component: ./path/to/local/module.json not found. Only json files are require-able.
```

I get the design decision of limiting local requires to `.json` files - keeping server.js small and limiting it's responsibility to component data fetching / plumbing 👍 .

However, the existence of a hard limitation in the framework seems a little too restrictive IMHO:
- A component can still be "small" and still have local requires
- Component authors can circumvent this restriction anyway by using a module bundler to generate `server.js` (in doing so bundle all of their local requires into one file)
- There isn't much difference between npm dependencies and local dependencies yet one is allowed and the other isn't.

My particular use case is as follows (React server side rendering):

```js
'use strict';

var React = require('react');
var ReactDOMServer = require('react-dom/server');

var Component = require('./path/to/my/component');

module.exports.data = function(context, callback) {
  var element = React.createElement(Component);
  var html = ReactDOMServer.renderToString(element);

  callback(null, { html: html });
};
```

Are there any major objections to this? Potential downsides I see are CLI bloat and percieved lack of control over `server.js` complexity. CLI bloat because we'd probably need some sort of module bundler to allow local requires. That being said we've discussed similar stuff in #339. And in terms of lack of control .... surely thats what code reviews are for 😇 .

matteofigus:
There were a couple of reasons for doing this a while ago but I agree that's not an issue anymore.

We use webpack here for doing this - I would vote 👍 for putting this into the CLI and leaving to webpack to produce the bundled server.js. I think this would be the easiest way to achieve this as we wouldn't change anything on the publishing and execution, only thing would be to add the extra magic to the CLI.

matteofigus:
Ah, there are other options apart from webpack obviously, we may do a spike to see what's the best way to do this :)

matthewdavidson:
Yeah I would have suggested webpack too!

matteofigus:
Also, @antwhite worked on this a while ago. I think we just may need to bring this code into the cli and add some testing: https://github.com/antwhite/oc-webpack

nickbalestra:
@matthewdavidson @matteofigus I like the idea of relying on a bundler + transpiler to solve this and https://github.com/opentable/oc/issues/339 issue. As per the goals I would suggest to go with rollup/bublé instead of the webpack/babel combo:

- rollup bundles are always smaller then browserify or webpack
[...]

Issue #339, opened 2017-01-03
ES2015 support in server.js


matthewdavidson:
Component authors are limited to writing `server.js` in es5 regardless of the node version the component registry is running on. I believe this is due to the usage of uglify (which [at present](https://github.com/mishoo/UglifyJS2/issues/448) supports only es5) in the component packaging process.

I see three options:

1. Introduce [babel](https://babeljs.io/docs/plugins/preset-es2015/) into the packaging process.
2. Drop uglify in favor of a minifier that supports es2015 (e.g. [babili](https://github.com/babel/babili) is the only one i know of 😒 )
3. Wait for uglify to [support es6](https://github.com/mishoo/UglifyJS2/issues/448)

Rationale: At best, it's awkward for component authors who write their client side code in es2015 to be limited to es5 in `server.js`. At worst it's confusing for newcomers to the codebase to see different javascript syntax across files.

Thoughts?

matteofigus:
Agree, this has been on my list like forever. I was waiting for uglify to catchup and resisting to add babel because it would make the cli to grow quite a lot - but perhaps it's now time to reconsider.

I didn't know babili but looking at the code it looks to me that it is anyways a combination of babel + uglify, so I think my preference would for short term doing 1. with `es2015` preset (as a pre-minification process) and then 3. long term when uglify will be ready.

matteofigus:
Done with #346

PR #344, opened 2017-01-08
Package server script cleanup

nickbalestra:
This initial cleanup is to facilitate adding webpack support to package-server-script in a follow-up PR as per #340 and #339


matteofigus:
Excellent, thanks @nickbalestra 

@matthewdavidson yes I think we are probably gonna totally get rid of that as the bundler should handle local requires nicely (both es6 imports and commonjs)

PR #347, opened 2017-01-11
342 - (OC Client) Perform GET instead of POST for single component

ivan-dimitrov-skyscanner:
Also:
- Extended get-components-data error handling to anticipate a flat response
  with a single error field when a major 4xx error is detected for the request.
  In this case we extend the error message to provide as much data as possible.
- Added more tests and fixed the existing ones.


ivan-dimitrov-skyscanner:
Hey @matteofigus, I have really no idea why it is failing. I don't think my changes should affect the failing test in any way :-(

Also one of the tests is showing green but instead of green tick it has three green '���' in front of it?!
Any ideas are welcome...

matteofigus:
Just hitting re-run worked ;)

ivan-dimitrov-skyscanner:
Awesome, thanks :-)

ivan-dimitrov-skyscanner:
Hi @matteofigus, can we merge this one as it becomes a kind of a blocker for us?
If there is something in particular that you don't like with this PR, just shout and we can refactor it :-)

test/acceptance/client.js
```diff
@@ -612,27 +654,15 @@ describe('The node.js OC client', function(){
         it('should contain the error details', function(){
 
           var expectedRequestWithExtraParams = {
-            url: 'http://localhost:1234',
-            method: 'post',
+            url: 'http://localhost:1234/hello-world/~1.0.0??hi=john',
```
matteofigus: Sorry, not sure I understand why there are 2 ?. What's wrong with 1?
Or, what about http://localhost:1234/hello-world/~1.0.0/?hi=john

client/src/href-builder.js
```diff
@@ -34,8 +34,8 @@ module.exports = function(config){
           registryUrl = clientRenderingEndpoint,
           registrySegment = registryUrl.slice(-1) === '/' ? registryUrl : (registryUrl + '/'),
           qs = !!component.parameters ? ('/?' + querystring.stringify(component.parameters)) : '';
-
-      return url.resolve(registrySegment, component.name) + versionSegment + qs;
+      
+      return url.resolve(registrySegment, component.name || '') + versionSegment + qs;
```
matteofigus: Can we actually validate component.name is not empty here and in case return a validation error? No point of making a req if component.name is missing.

client/src/get-components-data.js
```diff
@@ -5,11 +5,103 @@ var request = require('minimal-request');
...
+      callback(responses);
+    });
+  };
+
+  var performGet = function(endpoint, serverRendering, options, callback) {
+    var component = serverRendering.components[0];
+    var urlPath = component.name + (component.version ? '/' + component.version : '');
+    var queryString;
+
+    if (options.parameters) {
```
matteofigus: We have an `href-builder` for building the client-side failover urls. Can we move this there perhaps in a new "server" method?
https://github.com/opentable/oc/blob/master/client/src/href-builder.js#L9

I think you may be 1) able to reuse most of that stuff and perhaps refactor a bit as I think the only thing changing is the base url 2) keep the same convention which is version/?params instead of version??params

test/acceptance/client.js
```diff
@@ -38,7 +38,8 @@ describe('The node.js OC client', function(){
   var getRegExpFromJson = function(x){
     return JSON.stringify(x)
       .replace(/\+/g, '\\+')
-      .replace(/\[/g, '\\[');
+      .replace(/\[/g, '\\[')
+      .replace(/\?\?/g, '\\?'); //In our json regexp in order to preserve a single ? we are escaping it  with ??
```
matteofigus: Not a big fan I guess, see next comment where I ask for clarification

client/src/href-builder.js
ivan-dimitrov-skyscanner: Done!

test/acceptance/client.js
ivan-dimitrov-skyscanner: That's because expectedResults are actually regular expressions and '?' is a special symbol so we need to escape it. If we have an expected result like 'some ? text' we need to turn it into 'some \\? text' before the final processing (single back slash doesn't work because '\?' equals '?'). However we also run the getRegExpFromJson(x) function on it, which employs JSON.stringify() that in turn will additionally escape every '\' with '\\' and our string becomes 'some \\\\? text' internally.
That's why I decided to escape it with double question mark '??'.
If you are asking if we can escape '?' with '/?' that's possible, too.
Any other ideas are welcome? :-)

matteofigus: Ok, it this is only for escaping that it is fine. Perhaps, just to be 100% sure about what happens, we may create a test for a component with parameters and test the response has the correct href with the structure `component/version/?params` - that could be done without regex and would be useful for tests readability I think?

ivan-dimitrov-skyscanner: Yes, it is only for escaping the symbol in the regex and only in the tests. Regarding the second part - I'm not quite sure I understand what you are asking for...
If it's to add an additional test about if we set the right URL/href address then the only way this can be done in the client.js acceptance tests is to test for an error condition. There we are getting the request URL in the error message (exactly what's happening in the tests). However if we want to intercept and check the request then we should create a unit test for get-component-data because renderComponent/s methods don't us the full response, just the data in it.

test/acceptance/client.js
ivan-dimitrov-skyscanner: @matteofigus, I did some changes and introduced new tests to cover the expectations we have for get-component-data to prepare a GET request when just one component is passed and for href-builder to include the component parameters as URL query params for the GET request URL.
However I can't come up with an idea for getting rid of this '??' escaping. Perhaps I don't understand your question entirely :-(

matteofigus:
Thanks @ivan-dimitrov-skyscanner this is a great work!

PR #348, opened 2017-01-12
Remove `.only`

nicotaing:
@matteofigus 

I'm confused how it still runs the whole suite with `.only`. Anyway might as well to remove it.


matteofigus:
Ah! Good catch, thanks @nicotaing 

I think the "whole" is unit, integration and client-side, but in case of acceptance, it was executing only this one (each suite runs in a separate task and the only is relative to that).

PR #362, opened 2017-01-21
Node 7

matteofigus:
Just a test for now, definitely worth keeping an eye on it and integrating as soon as it gets green

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ matrix:
       env: SAUCE=true
     - node_js: "5"
     - node_js: "6"
+    - node_js: "7"
 
 before_script:
 - npm install -g grunt-cli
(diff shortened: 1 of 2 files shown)


matteofigus:
I think the main issue is node 7 comes with npm 4, where there is a breaking change that is breaking the tests (dep not found): `Partial shrinkwraps are no longer supported. npm-shrinkwrap.json is considered a complete installation manifest except for devDependencies.`

So I guess I'll try to rework the npm shrinkwrap in order to work with npm 4 /cc @nickbalestra

matteofigus:
All green, to you @nickbalestra for reviewing and merging. Not sure reviewing the shrinkwrap is the most exciting task in the planet. I guess you can just see that with node code changes node 7 gets green :rofl: 

So, basically, trust me, it works

Issue #358, opened 2017-01-20
ES6 component fails to compile when object properties shorthand 


matteofigus:
server.js is 
```js
'use strict';

module.exports.data = (context, callback) => {
  const username = context.params.username || 'matteo';
  callback(null, { username });
};
```
Related to #346 

This is running the latest version of OC (0.34.0) with node v4.7.2

```
an error happened while packaging /Users/mfigus/Documents/ot/test/es6: ./es6/server.js
Module build failed: SyntaxError: Unexpected token (7:28)
    at Parser.pp.raise (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:1745:13)
    at Parser.pp.unexpected (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:2264:8)
    at Parser.pp.parsePropertyValue (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:1344:15)
    at Parser.pp.parseObj (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:1311:10)
    at Parser.pp.parseExprAtom (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:1124:19)
    at Parser.pp.parseExprSubscripts (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:1023:19)
    at Parser.pp.parseMaybeUnary (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:1004:19)
    at Parser.pp.parseExprOps (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:948:19)
    at Parser.pp.parseMaybeConditional (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:930:19)
    at Parser.pp.parseMaybeAssign (/Users/mfigus/Documents/ot/oc/node_modules/falafel/node_modules/acorn/dist/acorn.js:908:19)
```

nickbalestra:
Thanks @matteofigus .
There are a couple of thing causing this, that I'm currently invesigating:

1. falafel-loader -> acorn 
2. webpack uglify plugin

[...]

PR #363, opened 2017-01-21
Extending Integration tests

nickbalestra:
Added complex integration test to cover situations like the one reported with #358

### DIFF
--- a/test/integration/cli-domain-package-server-script/cli-domain-package-server-script.js
+++ b/test/integration/cli-domain-package-server-script/cli-domain-package-server-script.js
@@ -3,6 +3,7 @@
 var expect = require('chai').expect;
 var fs = require('fs-extra');
 var path = require('path');
+var sinon = require('sinon');
 var packageServerScript = require('../../../src/cli/domain/package-server-script/index.js');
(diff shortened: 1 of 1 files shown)


test/integration/cli-domain-package-server-script/cli-domain-package-server-script.js
```diff
@@ -386,5 +387,77 @@ describe('cli : domain : package-server-script', function(){
...
+              done();
+            });
+
+            afterEach(function(done){
+              require.cache[path.resolve(publishPath, serverName)] = null;
+              done();
+            });
+
+            it('should save compiled data provider', function(done){
+              packageServerScript(
```
matteofigus: what if we bring all of this stuff inside the beforeEach and we save the result (compiledContent) into a variable, so then we can have 3 individual it sync tests that can explain a little bit better what we are doing? 
e.g. first test we have no error, then test wrapping correctly works for do, for, and while? I think it would improve readability and maintainability

nickbalestra: Good point!

nickbalestra:
@matteofigus updated the tests, let me know what you think

PR #364, opened 2017-01-22
Don't allow routes that contain prefix

### LINKED ISSUE #350: Config routes containing same prefix as components are never hit
Hi @matteofigus , just a small issue here
In my config, I have something like
```
prefix: '/abc/',
routes: [{
        route: '/abc/<some_custom_route>',
        method: 'get',
        handler: function(req, res) {
           //return data
        }
    }]
```
But this route is never hit as the routes added via config are registered after the default OC component routes
https://github.com/opentable/oc/blob/master/src/registry/router.js#L54

If you agree, more than happy to make a PR for a fix :)

matteofigus:
Fixes #350

### DIFF
--- a/src/registry/domain/validators/registry-configuration.js
+++ b/src/registry/domain/validators/registry-configuration.js
@@ -1,9 +1,10 @@
 'use strict';
 
+var format = require('stringformat');
 var _ = require('underscore');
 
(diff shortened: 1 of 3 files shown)


ArnoldZokas:
LGTM

PR #367, opened 2017-01-24
Always create new S3 client before use

jwhayes:


matteofigus:
LGTM.

Just to clarify to external users: this is the result of a chat we had about some possible connectivity issues between the api and the public internet (the aws api). It turns out that during some network issues, the api instance seems not to be capable of auto-recovering and from a quick investigation this seems to be due to the aws-sdk client we use. 

This is a little test to see if this fixes without negatively impacting on performance and robustness. In case it will be unsuccessful, we will rollback.

PR #370, opened 2017-01-24
Cleanup on headers handling

matteofigus:
While cleaning up a bit the code (I wanted to remove the `delete` as it may be not great for v8's optimisation, I had the chance to look again at all the tests and I found a couple of things that I'd like to change.

1) When doing the GET /component I think headers should be set as headers and not included in the response
2) When doing the POST / for the batch endpoint I think headers should not be set but included in the response. Still, I moved them from being in the `response.headers` to just `headers` (one level down). I think this allows us to have "response" to be exactly identical as a GET /component response.

I would be tempted to publish this as a bugfix instead of breaking change if you agree @ivan-dimitrov-skyscanner 

Also, we may need to revisit your PR #347 as I think if we have all the headers in the POST perhaps we don't need that?

More details inline.


src/registry/routes/component.js
```diff
@@ -20,12 +20,8 @@ module.exports = function(conf, repository){
         res.errorDetails = result.response.error;
       }
 
-      if (!_.isEmpty(result.response.headers)) {
-        res.set(result.response.headers);
-        
-        if (req.method === 'GET') {
-          delete result.response.headers;
-        }
```
matteofigus: This is not actually needed as `component.js` is used only by the GET endpoint. Also `delete` can be not great for performances so I revert the strategy (to not set it in the response and then to make it part of the POST individual result)

test/acceptance/registry.js
```diff
@@ -167,7 +167,7 @@ describe('registry', function(){
       it('should return the component with custom headers', function() {
         expect(result[0].response.version).to.equal('1.0.0');
         expect(result[0].response.name).to.equal('hello-world-custom-headers');
-        expect(result[0].response.headers).to.be.deep.equal({'cache-control': 'public max-age=3600', 'test-header': 'Test-Value'});
+        expect(result[0].headers).to.be.deep.equal({'cache-control': 'public max-age=3600', 'test-header': 'Test-Value'});
```
matteofigus: I actually think I didn't pay so much attention on the previous review as I was surprised to see this. What do you think @ivan-dimitrov-skyscanner ? Can we safely change this and consider it as a bugfix?

test/unit/registry-routes-component.js
```diff
@@ -453,8 +453,7 @@ describe('registry : routes : component', function(){
     });
 
     it('should set response headers', function() {
-      expect(response.headers).to.not.be.null;      
-      expect(response.headers['test-header']).to.equal('test-value');
+      expect(response.headers).to.be.undefined;
```
matteofigus: I think this tests were actually broken. The GET shouldn't have headers inside the response so this fixes it.

ivan-dimitrov-skyscanner: Hey @matteofigus, thanks for getting back. Can you give me one more day to revisit it as well.

matteofigus: No problem take your time ;)

test/acceptance/registry.js
ivan-dimitrov-skyscanner: @matteofigus It seems as a kind of invariant to me. However I tried to checkout your changes locally and play it them and for some reason I could not see how to do it :-(
And the main thing we'd liked to accomplish was:
- in case of GET the headers should be visible as normal HTTP response headers
- in case of POST they should not be set as HTTP response headers (because there is no point to do it) but to be present in the normal response body.
And from this point of view your change looks OK to me.

test/unit/registry-routes-component.js
ivan-dimitrov-skyscanner: You're right - those tests were broken :-(

test/acceptance/registry.js
matteofigus: Yeah agree. I think the purposed change here is to move headers one level down on the POST so that instead of having
```json
{   
  components: [{
    status: 200,
    response: {
      ...
      headers: { ... }
    }
  }]
}
```
we have 
```json
{   
  components: [{
    status: 200,
    response: {
      ...
    },
    headers: { ... }
  }]
}
```

ivan-dimitrov-skyscanner: Agree, and also I'm OK to have it as a bugfix instead of a breaking change :-)

ivan-dimitrov-skyscanner:
@matteofigus, 
Regarding PR #347 - I think we'll still need it because even though we'll have all the headers in the POST, the majority of the HTTP proxies (CDN, etc.) which stay on the request-response path don't cache responses of POST requests. For this reason I think we still need a GET request when dealing with just single component.

matteofigus:
Thanks, so let me merge this and then we'll review that one ;)

PR #372, opened 2017-01-26
[Registry] fallbackRegistryUrl configuration parameter

jankowiakmaria:
# What's this about?
This PR introduces new registry configuration parameter `fallbackRegistryUrl`. It allows developers to set another registry as a fallback in addition to their local one, so that they can focus on developing their single component within larger system.

### Registry -> configuration
```
|<sub>fallbackRegistryUrl</sub>|string (url)|no|Url to another registry which will be used by current registry when component cannot be found in current registry|
```
#### usage update:
```
oc dev <dirName> [port] [baseUrl] [--hotReloading=value [--fallbackRegistryUrl=value]]
```
#### parameters (table) update:
```
|fallbackRegistryUrl|Url to another registry which will be used by dev registry when component cannot be found in local registry. Example: --fallbackRegistryUrl=http://anotherhost:anotherport/||
```


matteofigus:
Excellent @pbazydlo and @jankowiakmaria - I just updated the wiki and published a new version.

PR #371, opened 2017-01-25
oc-template-handlebars module

nickbalestra:
- [x] To be reviewed/rebased once https://github.com/opentable/oc-template-handlebars v1.0.0 has been published

PR #375, opened 2017-01-31
oc-template-jade module

nickbalestra:
Done per jade what we did for handlebars in #371 
To be rerun once oc-template-jade will be published

### DIFF
--- a/src/cli/domain/package-template.js
+++ b/src/cli/domain/package-template.js
@@ -11,23 +11,24 @@ var hashBuilder = require('../../utils/hash-builder');
 var strings = require('../../resources');
 var validator = require('../../registry/domain/validators');
 
+var templateEngines = {
+  handlebars,
+  jade
+};
+
 var javaScriptizeTemplate = function(functionName, data){
   return format('var {0}={0}||{};{0}.components={0}.components||{};{0}.components[\'{1}\']={2}', 'oc', functionName, data.toString());
 };
 
 var compileView = function(viewPath, type) {
-    var template = fs.readFileSync(viewPath).toString(),
-        preCompiledView;
+  var template = fs.readFileSync(viewPath).toString();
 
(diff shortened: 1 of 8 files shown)


src/cli/domain/package-template.js
```diff
@@ -20,11 +20,7 @@ var compileView = function(viewPath, type) {
         preCompiledView;
 
   if(type === 'jade'){
-    preCompiledView = jade.compileClient(template, {
-      filename: viewPath,
-      compileDebug: false,
-      name: 't'
-    }).toString().replace('function t(locals) {', 'function(locals){');
+    preCompiledView = jade.precompile(template, { filename: viewPath });
```
matteofigus: I think I would like to do something similar to the oc-client here rather than all the ifs. Something like

```js
var precompilers = { jade, handlebars};
// and then
if(!precompilers[type]){
  throw strings.errors.cli.TEMPLATE_TYPE_NOT_VALID;
}
var preCompiledView = precompilers[type].precompile(template, { viewPath });
```

matteofigus: Also, note that in jade we named that viewPath, not filename 👍

src/registry/routes/helpers/get-component.js
```diff
@@ -235,9 +236,7 @@ module.exports = function(conf, repository){
               var template;
 
               if (component.oc.files.template.type === 'jade') {
-                var context = { jade: require('jade/runtime.js')};
-                vm.runInNewContext(templateText, context);
-                template = context.oc.components[key];
+                template = jade.getPrecompiledTemplate(templateText, key);
```
matteofigus: same as before here.

nickbalestra:
Updated.

nickbalestra:
Should be fixed now

PR #376, opened 2017-02-01
Cleanup un-needed files

matteofigus:
We forgot this on previous PR #375


client/src/get-compiled-template.js
```diff
@@ -1,14 +1,20 @@
...
+var jade = require('oc-template-jade');
 var request = require('minimal-request');
 
-var executor = require('./executor');
 var settings = require('./settings');
 var TryGetCached = require('./try-get-cached');
 
 module.exports = function(cache){
 
+  var renderers = {
```
nickbalestra: Used templateEngines instead of renderers, I think will be better if we try to stick to the same name convention

matteofigus: done

PR #381, opened 2017-02-13
removed define production on server bundle

nickbalestra:
As reported by @tkant 

```
NODE_ENV=test oc dev . 3030
console.log(’NODE_ENV: ‘ + process.env.NODE_ENV);
Its throwing 'production'
```

### DIFF
--- a/src/cli/domain/package-server-script/bundle/config/index.js
+++ b/src/cli/domain/package-server-script/bundle/config/index.js
@@ -42,11 +42,6 @@ module.exports = function webpackConfigGenerator(params){
         }
       ]
     },
-    plugins: [
-      new webpack.DefinePlugin({
-        'process.env.NODE_ENV': JSON.stringify('production')
-      })
-    ],
     resolveLoader: {
       modules: ['node_modules', path.resolve(__dirname, '../../../../../../node_modules')]
     }


src/cli/domain/package-server-script/bundle/config/index.js
```diff
@@ -42,11 +42,6 @@ module.exports = function webpackConfigGenerator(params){
         }
       ]
     },
-    plugins: [
```
tkant: I wonder why we needed this in first place :thinking:

matthewdavidson:
Bear in mind that this will leave all of react's dev warnings in place (see https://facebook.github.io/react/docs/installation.html#development-and-production-versions). 

Given that `oc-registry` processes *should* be executed with `NODE_ENV=production` then it shouldn't really matter. Do we have docs around best practices for running the registry? Perhaps an example `Dockerfile`?

nickbalestra:
@tkant that's was one of the reason we had this in place (see @matthewdavidson comment). I think will be better to keep it as it is, until we properly deal with react and co. (Not sure that this will happen within the server.js part, but I can see that for now that's the only way around)

matteofigus:
It is not recommended to access process from the component. This is why the registry maintainer can set the env on the registry configuration and make it accessible from the context: https://github.com/opentable/oc/wiki/Registry#setup

tkant:
Thats right, but don't you think that injecting context here and there in the code makes it less maintainable. We should always try to make things modular. Just a suggestion :)

tkant:
Or is there any other way we can get the registry's configuration parameter in any of the open-component available?

matteofigus:
@tkant 

Basically, the registry maintainer makes this information available on each instance, so that in your component you'll be able to access it. Only difference is instead of accessing it via process.env you'll get it via context.env.  

```js
// registry

var registry = new oc.registry({
  env: { name: "production", otherStuff: "hello" }
  ...
});

registry.start(...);

// your component
module.exports.data = (ctx, callback) => {
  const isProd = ctx.env.name === 'production';
  ...
}
```

tkant:
@matteofigus I understood the context thing in the first place. The way I had wrote our OC is we are requiring config.js at top of many module. So my data provider have them on top of server.js. 
One solution is to create a global var. But I really don't like polluting global space. :(

matteofigus:
Ok, maybe I find difficult to understand as I design the components differently. Can you share a component example for letting us understand better your use-case? Thanks

tkant:
Ohkk.. Here is how I had done. 
Divided my modules based on feature we have. Those features uses config var to make appropriate backend calls. So my server.js (which is data provier) its like this:
```
'use strict';

const config = require('./configs').getConfig();
const bootstrap = require('./app/bootstrap');
... // other helpers

module.exports.data = function (context, callback) {
 // we can get context.env here but how can I use it in other modules which I had included above.
 // 1 option is to call require inside this and assign it to some global variable. For that as well I need to 
 // move all require statements which utilizes my config. 
}
```

For config I had wrote this:
```
'use strict';

const env = require('./env.json');

module.exports.config = function () {
  var nodeEnv = process.env.NODE_ENV || 'test'; // here we can initialize the nodeEnv with global var
  return env[nodeEnv];
};
```
env.json contains something like this:
```
{
  “test”: {}
  “production”: {}
}
```

tkant:
So I can also move the require statements in some other JS and include it inside module.exports.data

tkant:
@matteofigus I would love to see how you wrote OCs.. can you point me to some repo ?

matteofigus:
Ah! I see.
The way we do it is very similar but slightly different. We still have something like `env.json`. The difference is, we do all the requires first, then we get the right stuff inside the context. So basically everything you require exposes then a constructor where you can pass the context through.

Simple mode, the way we usually do this:
```js
const config = require('./config.json');
module.exports.data = (context, callback) => {
  var settings = config[context.env.name];
  ...
}
```
Or you can still have your js wrapper but you need to require it first and then initialise it later:

```js
const config = require('./config.js');
module.exports.data = (context, callback) => {
  var settings = config.getConfig(context.env);
  ...
}
```

You are right, I think an example would be good for this scenario. I definitely can put something on the wiki 👍

tkant:
Thanks @matteofigus 😃

PR #389, opened 2017-02-24
Fix regex on win paths normalisation

### LINKED ISSUE #386: oc publish using version 0.34.8 to an S3 bucket ignoring the folder structure
oc version - 0.34.8
os - Windows 10
I was trying to do an oc publish to an S3 bucket. My project gets published, but without any folder structure and all the file names are modified to include the folder names with backslash. So I get a flat structure for my project due to which the resources being referenced from my files becomes unavailable. This issue has been observed on MAC too.

matteofigus:
Fixes #386 

After doing some hacking, I found that on windows the path was normalised from `\\something\\something\\something` to `/something\something\something` instead of `/something/something/something`. Then I had a better look at the normalisation regex and spotted the issue.

### DIFF
--- a/src/registry/domain/s3.js
+++ b/src/registry/domain/s3.js
@@ -109,7 +109,7 @@ module.exports = function(conf){
 
         async.each(files, function(file, cb){
           var relativeFile = file.substr(dirInput.length),
-              url = (dirOutput + relativeFile).replace(/\\/, '/');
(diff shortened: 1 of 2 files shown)


mattiaerre:
never forget the **g**lobal search!
this LGTM ... shall I merge it now?

matteofigus:
sure, thanks

PR #393, opened 2017-03-03
Tweaking README

matteofigus:
1) I like serverless architectures, but I feel like it is half loved half hated. Also, most often refers to frameworks on top of lambda, and oc is not. So I think we should remove it.
2) A recurrent theme when we made the OC meetup was about iframes and how much of OC was inspired by iframes and how they should be in an ideal world.

What do you think about this little changes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,12 @@
 ![oc](https://raw.githubusercontent.com/opentable/oc/master/logo.png)
 =============
 
-OpenComponents, **serverless in the front-end world**.
-
 OpenComponents is an open-source framework that allows fast-moving teams to easily build and deploy front-end components. It abstracts away complicated infrastructure and leaves developers with very simple, but powerful building blocks that handle scale transparently.
+
+Imagine something **as simple as an iframe, that is not an actual iframe**:
+```html
+<oc-component href="//oc-registry.com/hello/1.x.x?name=John"></oc-component>
+```
 
 #### How does it work?
 


mattiaerre:
@matteofigus 
I personally like the Serverless concept and to me, OC fits exactly in this context given that S- doesn't mean "no server" but it means to focus on your application (aka the business) while someone else will help you w/ the infrastructure. But I do agree that unfortunately Amazon w/ Lambda is more or less what people nowadays associate w/ S-.
On top of that, I don't like very much the `iframe` approach as it reminds me a very old concept of the web if you know what I mean and I fear that this can be a bad "marketing strategy" for OC.
I do agree that we should rephrase this paragraph but I would like to put the emphasis more on the "deployment" and "UI" parts of this platform; I would try to use some of these keywords:

- FE
- Deploy
- Platform
- Service
- Distribute
- Publish
- Share
- UI
- Widget
- _aaS
- Element
- Interface
- Components

And by the way, why "Open"? Open for what? Maybe it's in there the real meaning of this beautiful thing that we love.

P.S. too long I know and, as usual, more questions than answers but I like this kind of conversations ... 😅

matteofigus:
Very very good ideas and very good points, thanks to both. I am closing this but we can keep talking about it here if we want.

@mattiaerre now that I see your keywords, specifically deployment, I got back a memory from Kevin's talk, a slide that I loved had the title "The joy of deployments". Perhaps that's something we could mention 👍

PR #396, opened 2017-03-08
Make oc-client.js supporting more templates via an extendable API

nickbalestra:
Make oc-client.js supporting more templates other then jade and handlebars (oc-template-react,..) through an API that allows to extend such support.

***

Context: The Api need to be able to register the new type of template supported and its various requirements (dependecies,..) Internally the architecture should take care of hadnling all the aspects needed in order to be able to render the new type of components (i.e: oc-template-react).

pre-configuration API:
======================
```javascript
<script>var oc = { conf: { templates: [] }};</script>
<script src="//registry.components.com/oc-client/client.js"></script>
```

post-configuration API:
=======================
```javascript
<script src="//registry.components.com/oc-client/client.js"></script>
<script>oc.registerTemplates()</script> 
```

- Render after a new template has been registered

### DIFF
--- a/src/components/oc-client/src/oc-client.js
+++ b/src/components/oc-client/src/oc-client.js
@@ -32,8 +32,6 @@ var oc = oc || {};
   // Constants
   var CDNJS_BASEURL = 'https://cdnjs.cloudflare.com/ajax/libs/',
       IE9_AJAX_POLYFILL_URL = CDNJS_BASEURL + 'jquery-ajaxtransport-xdomainrequest/1.0.3/jquery.xdomainrequest.min.js',
-      HANDLEBARS_URL = CDNJS_BASEURL + 'handlebars.js/4.0.5/handlebars.runtime.min.js',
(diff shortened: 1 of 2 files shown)


src/components/oc-client/src/oc-client.js
```diff
@@ -294,25 +310,35 @@ var oc = oc || {};
 
   oc.render = function(compiledViewInfo, model, callback){
     oc.ready(function(){
-      if(!!compiledViewInfo.type.match(/jade|handlebars/g)){
+
+      var template = templates.find(function(template){return template.type === compiledViewInfo.type});
```
matteofigus: find is not supported in IE: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find
Perhaps we can use something available in jQuery?

nickbalestra: I'll go vanilla, no need for a $ dependency just for that in here imho.

matteofigus:
This is fantastic, thanks @nickbalestra

PR #399, opened 2017-03-09
Changed option parser for cli

### LINKED ISSUE #135: Change option parser for cli
as the one currently used (https://github.com/harthur/nomnom):
- is not maintained: https://github.com/harthur/nomnom#deprecation-notice
- has a lot of bugs

I can work on it - any recommendations for parser?
https://github.com/tj/commander.js?

jankowiakmaria:
resolves #135 

Differences:
- removed `version` command (it has `version` option instead)
- no description of positional arguments
- added examples for each command (which should mitigate ^^)
- positional argument !== option (in nomnom both types could be used interchangeably)
- producing completion script is done differently (don't need additional dependencies - yargs provides completion script out of the box) - so wiki needs to be updated

~~Question - should I shrinkwrap or is it done during publishing/at some other point of the process?~~


src/cli/commands.js
```diff
@@ -2,92 +2,95 @@
 'use strict';
 
 module.exports = {
-  oc: {
+  usage: 'Usage: $0 <command> [options]',
+  commands: {
 
     dev: {
-      help: 'Runs a local oc test registry in order to develop and test components',
+      cmd: 'dev <dirName> [port] [baseUrl]',   //should be dirPath imho
```
nickbalestra: agree, it should be dirPath

src/cli/commands.js
```diff
@@ -2,92 +2,108 @@
...
+          default: true
         }
-      }
+      },
+      usage: 'Usage: $0 dev <dirName> [port] [baseUrl] [options]'
     },
 
     init: {
-      help: 'Creates a new empty component in the current folder',
+      cmd: 'init <componentName>',
```
matteofigus: shouldn't this be `init <componentName> [templateType]` ?

src/cli/facade/registry.js
```diff
@@ -12,57 +12,9 @@ module.exports = function(dependencies){
...
-        if(err){
-          log.err(err);
-          return callback(err);
-        }
 
-        log.ok(strings.messages.cli.REGISTRY_REMOVED);
-        callback(null, 'ok');
-      });
-    }
+    callback(null, 'ok');
```
matteofigus: should this file just be removed completely?

src/cli/facade/version.js
```diff
@@ -1,9 +0,0 @@
-'use strict';
```
matteofigus: I guess the reason is that the module supports a version command natively?

package.json
```diff
@@ -89,6 +87,7 @@
     "uglify-js": "2.6.4",
     "underscore": "1.8.3",
     "watch": "0.19.1",
-    "webpack": "2.2.0"
+    "webpack": "2.2.0",
+    "yargs": "^6.6.0"
```
matteofigus: let's pair on the shrinkwrap

src/cli/facade/registry.js
matteofigus: or, should it return a failing callback? (I am not sure I completely remember how this works)

src/cli/commands.js
```diff
@@ -2,92 +2,108 @@
...
-        },
         fallbackRegistryUrl: {
-          help: 'Url to another registry which will be used by dev registry when component cannot be found in local registry. Example: --fallbackRegistryUrl=http://anotherhost:anotherport/',
-          required: false
+          description: 'Url to another registry which will be used by dev registry when component cannot be found in local registry',
+        },
+        hotReloading: {
+          boolean: true,
+          description: 'Enables hot reloading. Note: when hot reloading is set to true, each request to the component will make the registry to create a new instance for the javascript closures to be loaded, while when false the instance will be recycled between components executions',
+          default: true
```
jankowiakmaria: As I've mentioned in PR comment - we can either have a positional argument or an option - yargs is really good but it offers different set of features than nomnom (it currently doesn't support description of positional arguments - I've seen workarounds to support it but they're not great)

src/cli/commands.js
jankowiakmaria: as above - if you'd like to I can walk you through differences (I've tried to include everything in the PR comment).

src/cli/commands.js
matteofigus: Sorry I totally forgot to read your first PR description. This is fine, sorry :)

src/cli/commands.js
matteofigus: templateType used to be a positional arg https://github.com/opentable/oc/wiki/Cli#init so I think I would like to keep it as is (here you are basically changing it to be an option)

package.json
matteofigus: Done

matteofigus:
@jankowiakmaria ok all good, only thing is on the oc init

src/cli/commands.js
jankowiakmaria: fine for me :) will change it
in nomnom you could use it as either [positional argument/option] of them - it would work anyway.

matteofigus:
Amazing! Are you able to update the wiki @jankowiakmaria ?

jankowiakmaria:
yep, sure :)

jankowiakmaria:
@matteofigus updated https://github.com/opentable/oc/wiki/Cli

PR #400, opened 2017-03-09
Fixing typo on registry UI for filtering.

ryaanwells:
Current the check boxes to hide deprecated and experimental components are broken on Safari due to a missing close bracket. This somehow works fine in Chrome.


matteofigus:
Ah! Good spot thanks for the fix @ryaanwells

PR #401, opened 2017-03-10
Webpack verbose

### LINKED ISSUE #377: --verbose flag
At the moment, when packaging the data provider webpack will output some info to the terminal following the setting:

```
stats: {
    chunks: false,
    colors: true,
    version: false,
    hash: false
  }
```

Although pretty minimal, I can see how this could be annoying when working on data providers that simply deal with jade/handlebar templateEngines.
That output is instead very useful when working on modern view provider that will be handled by react or similar templateEngines... Because of this, I suggest we add a `--verbose` flag option on the `dev` command to output extra bundling info while muting it by default (`stats: 'errors-only'`). This flag could be then also passed to the templateEngine...

matteofigus:
Fixes #377

I tested it locally and seems to work nicely :)
Can you have a look at it @nickbalestra?

### DIFF
--- a/src/cli/domain/package-components.js
+++ b/src/cli/domain/package-components.js
@@ -12,12 +12,10 @@ var getUnixUtcTimestamp = require('../../utils/get-unix-utc-timestamp');
 var validator = require('../../registry/domain/validators');
 
 module.exports = function(){
-  return function(componentPath, minify, callback){
+  return function(options, callback){
 
-    if(_.isFunction(minify)){
-      callback = minify;
-      minify = true;
-    }
+    var componentPath = options.componentPath;
+    var minify = options.minify || true;
 
     var files = fs.readdirSync(componentPath),
         publishPath = path.join(componentPath, '_package');
@@ -72,7 +70,8 @@ module.exports = function(){
           componentPath: componentPath,
           dependencies: component.dependencies,
           ocOptions: component.oc,
-          publishPath: publishPath
+          publishPath: publishPath,
+          verbose: options.verbose
         }, function(err, packagedServerScriptInfo){
           if(err){ return cb(err); }
 
--- a/src/cli/domain/package-server-script/index.js
+++ b/src/cli/domain/package-server-script/index.js
@@ -6,21 +6,10 @@ var path = require('path');
 var bundle = require('./bundle');
 var hashBuilder = require('../../../utils/hash-builder');
 
-var webpackDefault =  { stats: 'errors-only' };
-
-var webpackVerbose =  {
-  stats: {
-    chunks: false,
-    colors: true,
-    version: false,
-    hash: false
-  }
-};
-
 module.exports = function packageServerScript(params, callback){
   var fileName = 'server.js';
   var publishPath = params.publishPath;
-  var webpackParams = params.verbose ? webpackVerbose : webpackDefault;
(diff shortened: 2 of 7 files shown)


src/cli/domain/package-server-script/index.js
```diff
@@ -2,10 +2,13 @@
...
-var hashBuilder = require('../../../utils/hash-builder');
+
 var bundle = require('./bundle');
+var hashBuilder = require('../../../utils/hash-builder');
+
+var webpackDefault =  { stats: 'errors-only' };
 
-var webpackDefaults =  {
+var webpackVerbose =  {
   stats: {
```
nickbalestra: Now that verbosity is hidden behind a flag, It might be better to rely on some more verbose [presets that webpack](https://github.com/webpack/docs/wiki/node.js-api#statstojsonoptions) offers.
What about `stats: "verbose"` or`stats: "normal"`?

src/cli/domain/package-server-script/index.js
```diff
@@ -17,9 +20,10 @@ var webpackDefaults =  {
 module.exports = function packageServerScript(params, callback){
   var fileName = 'server.js';
   var publishPath = params.publishPath;
+  var webpackParams = params.verbose ? webpackVerbose : webpackDefault;
```
nickbalestra: Why not doing all this logic at the time you define the webpackParams?

i.e: 

```js
webpackParams = {
  stats: params.verbose ? 'verbose' : 'errors-only'
}
```

src/cli/facade/dev.js
```diff
@@ -75,7 +75,14 @@ module.exports = function(dependencies){
         log.warn(strings.messages.cli.PACKAGING_COMPONENTS, true);
 
         async.eachSeries(componentsDirs, function(dir, cb){
-          local.package(dir, false, function(err){
+
+          var packageOptions = {
+            componentPath: dir,
+            minify: false,
```
nickbalestra: What does `minify` do ?

matteofigus: minification of js/css static assets

src/cli/domain/package-components.js
```diff
@@ -12,12 +12,10 @@ var getUnixUtcTimestamp = require('../../utils/get-unix-utc-timestamp');
 var validator = require('../../registry/domain/validators');
 
 module.exports = function(){
-  return function(componentPath, minify, callback){
```
matteofigus: I changed this signature to accept an options parameter instead of adding another one

src/cli/facade/dev.js
nickbalestra: Isn't the PR for webpack verbosity? Or the PR scope changed?

matteofigus: Correct. I changed the signature there, not altering the behaviour. The false there was for minify

src/cli/domain/package-components.js
matteofigus: @nickbalestra ^

nickbalestra: I was wondering if we should have a minify option at all, but probably I'm missing some context. Why would we want to set minify, what are we trying to solve?

matteofigus: Minify is been there for a long time. It is not referred to server.js, not to template.js, but to actual static contents in the static folders. It is set component based on the package.json, and we use it for some components at opentable. I can show you some examples on Monday.

PR #415, opened 2017-03-22
[GPT-526] Dynamic requires of oc-templates

nickbalestra:


client/src/get-compiled-template.js
```diff
@@ -32,11 +23,20 @@ module.exports = function(cache){
...
-        var type = template.type;
-        if (type === 'jade') { type = 'oc-template-jade'; }
-        if (type === 'handlebars') { type = 'oc-template-handlebars'; }
 
-        cb(null, templateEngines[type].getCompiledTemplate(templateText, template.key));
+        var type = template.type;
+        var ocTemplate;
+        try {
+          if (type === 'jade') { type = 'oc-template-jade'; }
+          if (type === 'handlebars') { type = 'oc-template-handlebars'; }
```
matteofigus: can we move this ifs out of the try so it's clear the fail can happen only on the require line

client/src/get-compiled-template.js
```diff
@@ -32,11 +23,20 @@ module.exports = function(cache){
...
+
+          // dynamically require specific oc-template
```
matteofigus: I would remove this comment. If this is the only thing in the try, I think it's quite clear what's happening here.

client/src/get-compiled-template.js
matteofigus: can we move these ifs out of the try so it's clear the fail can happen only on the require line

client/src/get-compiled-template.js
```diff
@@ -32,11 +23,20 @@ module.exports = function(cache){
...
+          ocTemplate = require(type);
+        } catch (err) {
+          throw format(settings.templateNotSupported, type);
+        }
+        
+        cb(null, ocTemplate.getCompiledTemplate(templateText, template.key));
```
matteofigus: I am thinking that these 5 lines of code are actually repeated 4 times. Given we need them for both npm modules (oc and oc-client) - what about moving it to its own npm module, like `oc-require-templates` so that we can just require it all the times? If that doesn't require too much time obviously?
If not, I would just move it to its own file, like `oc-require-templates.js` and duplicate it inside the /client and /src folders.

nickbalestra: I would go with #2 for the moment so that we don't spread around too many pieces. Once out of 'flux', this could definitely go in its own npm module.

nickbalestra:
@matteofigus all points should be addressed

PR #418, opened 2017-03-24
Transpile static js before minification

### LINKED ISSUE #409: Static js needs to be transpiled before minification
As we don't transpile to es2015 completely, uglify will trow. 
Proposed solution: swap uglify with [babili](https://github.com/babel/babili)

matteofigus:
Fixes #409 

Quick note, I think we discussed about using babili, but after having a chat with @nickbalestra and making a test, I decided to still use uglify, but with babel on top.

With a test I made (using moment.js), babel+uglify => ~30KB VS babel+babili => ~45KB - so it turns out es5 + uglify is still more effective. We obviously may revisit in the future if makes sense.

### DIFF
--- a/src/cli/domain/package-static-files.js
+++ b/src/cli/domain/package-static-files.js
@@ -1,6 +1,8 @@
 'use strict';
 
 var async = require('async');
+var babel = require('babel-core');
+var babelPresetEnv = require('babel-preset-env');
 var CleanCss = require('clean-css');
 var format = require('stringformat');
 var fs = require('fs-extra');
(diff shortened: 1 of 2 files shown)


char1e5:
@matteofigus @nickbalestra This version `0.36.12` breaks our widget static bundle. 
We still stuck in early es5. https://github.com/opentable/reservation-widget
screen shot 2017-03-24 at 11 39 11 am

matteofigus:
@char1e5 yes, while we fix it, oc.minify = true on the package.json
https://github.com/opentable/oc/wiki/package.json-structure

PR #421, opened 2017-03-25
Minify static content by transpiling it first

### LINKED ISSUE #409: Static js needs to be transpiled before minification

matteofigus:
Fixes #409 

Summary: #418 was opened and merge to fix #409 but we had to rollback as the packaging was producing an invalid bundle (not good for browsers and potentially fragile for uglify). So #420 reverted it. This should be finally the one, with more testing, more fixes etc.

To review, I recommend to go commit by commit.

### DIFF
--- a/src/cli/domain/package-static-files/index.js
+++ b/src/cli/domain/package-static-files/index.js
@@ -34,7 +34,7 @@ var copyDir = function(params, cb){
 
         if(params.minify && params.ocOptions.minify !== false && (fileExt === '.js' || fileExt === '.css')){
           var fileContent = fs.readFileSync(filePath).toString(),
-              minified = minifyFile(fileExt, fileContent, params.ocOptions);
+              minified = minifyFile(fileExt, fileContent);
 
           fs.writeFileSync(fileDestination, minified);
         } else {
--- a/src/cli/domain/package-static-files/index.js
+++ b/src/cli/domain/package-static-files/index.js
@@ -1,41 +1,14 @@
 'use strict';
 
 var async = require('async');
-var babel = require('babel-core');
-var babelPresetEnv = require('babel-preset-env');
-var CleanCss = require('clean-css');
 var format = require('stringformat');
 var fs = require('fs-extra');
+var minifyFile = require('./minify-file');
 var nodeDir = require('node-dir');
 var path = require('path');
-var uglifyJs = require('uglify-js');
 var _ = require('underscore');
 
-var strings = require('../../resources');
-
-var minifyFile = function(fileType, fileContent, ocOptions){
-
-  if(fileType === '.js'){
-
-    var presetOptions = {
-      targets: {
-        browsers: 'ie 8',
-        uglify: true
-      },
-      useBuiltIns: true,
-      modules: false
-    };
-
-    var babelOptions = { presets: [[babelPresetEnv, presetOptions]] },
-        es5 = babel.transform(fileContent, babelOptions).code;
-    
-    return uglifyJs.minify(es5, { fromString: true }).code;
-  } else if(fileType === '.css'){
-    return new CleanCss().minify(fileContent).styles;
-  }
-
-  return fileContent;
-};
+var strings = require('../../../resources');
 
 var copyDir = function(params, cb){
   var staticPath = path.join(params.componentPath, params.staticDir),
--- a/src/cli/domain/package-static-files/minify-file.js
+++ b/src/cli/domain/package-static-files/minify-file.js
@@ -5,7 +5,7 @@ var babelPresetEnv = require('babel-preset-env');
 var CleanCss = require('clean-css');
 var uglifyJs = require('uglify-js');
 
-module.exports = function(fileType, fileContent, ocOptions){
+module.exports = function(fileType, fileContent){
 
   if(fileType === '.js'){
 
@@ -24,7 +24,7 @@ module.exports = function(fileType, fileContent, ocOptions){
     return uglifyJs.minify(es5, { fromString: true }).code;
 
   } else if(fileType === '.css'){
-    
+
     return new CleanCss().minify(fileContent).styles;
   }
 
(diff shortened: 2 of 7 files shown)


src/cli/domain/package-static-files/minify-file.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+var babel = require('babel-core');
+var babelPresetEnv = require('babel-preset-env');
+var CleanCss = require('clean-css');
+var uglifyJs = require('uglify-js');
+
+module.exports = function(fileType, fileContent){
```
mattiaerre: this should be `fileExt` correct?

test/integration/cli-domain-package-static-files-minify-file.js
```diff
@@ -0,0 +1,39 @@
...
+		describe('when file contains es6', function(){
+
+			var content = 'const hi = (name) => `hello ${name}`;';
+			var minified;
+
+			before(function(){
+				minified = minifyFile('.js', content);
+			});
+
+			it('should minify it', function(){
```
mattiaerre: a snapshot test, in this case, would simplify what you need to assert in order to make sure that the minification happened.

src/cli/domain/package-static-files/index.js
```diff
@@ -64,7 +34,7 @@ var copyDir = function(params, cb){
 
         if(params.minify && params.ocOptions.minify !== false && (fileExt === '.js' || fileExt === '.css')){
           var fileContent = fs.readFileSync(filePath).toString(),
-              minified = minifyFile(fileExt, fileContent, params.ocOptions);
+              minified = minifyFile(fileExt, fileContent);
```
mattiaerre: this is the only consumer of `minifyFile` I presume.

src/cli/domain/package-static-files/minify-file.js
```diff
@@ -0,0 +1,32 @@
...
+        es5 = babel.transform(fileContent, babelOptions).code;
+    
+    return uglifyJs.minify(es5, { fromString: true }).code;
+
+  } else if(fileType === '.css'){
+
+    return new CleanCss().minify(fileContent).styles;
+  }
+
+  return fileContent;
```
mattiaerre: you shouldn't need to return here as the file extension check is done also at a higher level.

test/unit/cli-domain-package-components.js
```diff
@@ -6,96 +6,146 @@ var path = require('path');
...
+      join: path.join,
+      resolve: function(){
+        return _.toArray(arguments).join('/');
       }
-    },
-    path: pathMock,
-    './package-static-files': sinon.stub().yields(null, 'ok'),
-    './package-template': sinon.stub().yields(null, { type: 'jade', src: 'template.js', hashKey: '123456'})
-  }, { __dirname: '' });
+    };
```
mattiaerre: I see a lot of `arrange` before running this unit test. maybe this component does too many things?

test/unit/cli-domain-package-static-files.js
```diff
@@ -6,69 +6,75 @@ var path = require('path');
...
-    'node-dir': { paths: sinon.stub().yields(null, { files:[] })},
-    path: {
-      basename: path.basename,
-      dirname: function(){
-        return path.dirname.apply(this, _.toArray(arguments)).replace(/\\/g, '/');
-      },
-      extname: path.extname,
-      join: function(){
-        return path.join.apply(this, _.toArray(arguments)).replace(/\\/g, '/');
+describe('cli : domain : package-static-files', function(){
```
mattiaerre: same as above re the `arrange` phase.

test/unit/cli-domain-package-components.js
matteofigus: Agree. But I would separate in a different PR if you don't mind

test/unit/cli-domain-package-static-files.js
matteofigus: Agree. That's why I am starting to take one piece off (minify part). Definitely gonna tackle other little pieces but I would do in separate PRs if you don't mind

src/cli/domain/package-static-files/index.js
matteofigus: Yes

test/integration/cli-domain-package-static-files-minify-file.js
matteofigus: With a snapshot I would test that
```js
const hi = (name) => `hello ${name}`;
```
would always translate to
```js
var hi=function(n){return"hello "+n};
```
Which is fine in this scenario, but what if at some point I upgrade babel and this starts translating to 
```js
var hi=function(a){return"hello "+a};
```
So, my thinking was that I wanted to test something without taking the risk of coupling it too much to the snapshot. So I guess that was by design. Perhaps I'm overthinking this - as I see the benefit of seeing the simplicity of A to B transformation for a test readability point of view.

nickbalestra: I think that the process is referentially transparent , so it should return always the same result

matteofigus: ok, so let me update this.

test/unit/cli-domain-package-components.js
mattiaerre: Sure, happy to pair on this if you like.

test/integration/cli-domain-package-static-files-minify-file.js
mattiaerre: I personally think we should not test how the internal minification library works; that's why I was proposing a snapshot test. You will end up having a file containing your uglified code in one place. which is more or less what you do have now. What about a pairing session w/ you after this PR has been merged?

mattiaerre:
@matteofigus this LGTM, happy to merge if you like.

PR #423, opened 2017-03-27
[GPT-517] Exposing registry supported templates via context

nickbalestra:
- [x]  CLi dev configure with `oc-template-jade` and `oc-template-handlebars` by default
- [x] `getTemplate()` helper for the repository to leverage on template.getInfo() API
- [x] add templates to contextObj

@matteofigus did I miss anything? (Still WIP)

### DIFF
--- a/src/registry/domain/repository.js
+++ b/src/registry/domain/repository.js
@@ -175,6 +175,16 @@ module.exports = function(conf){
     getStaticFilePath: function(componentName, componentVersion, filePath){
       return this.getComponentPath(componentName, componentVersion) + (conf.local ? settings.registry.localStaticRedirectorPath : '') + filePath;
     },
+    getTemplates: function(){
+      return conf.templates.map(function(template){
+        var info = require(template).getInfo();
+        return {
+          type: info.type,
+          version: info.version,
+          externals: info.externals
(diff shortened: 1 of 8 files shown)


nickbalestra:
Want to add some tests for this

src/registry/domain/repository.js
```diff
@@ -175,6 +175,19 @@ module.exports = function(conf){
...
+          var info = require(template).getInfo();
+          return {
+            type: info.type,
+            version: info.version,
+            externals: info.externals
+          };
+        });
+      }
+      return [];
+    },
```
matteofigus: This is good, but I am thinking performance here. This if+map+getInfo() is executed on each component execution (when the context is baked) by making this call and I don't think there is actually any need for that, as the templates are kind of statically initialised during registry's start anyways. What about moving this to the constructor on the top (around line 17) and memorising that in a variable, so that this method just exposes it?

nickbalestra: totally agree!

nickbalestra:
@matteofigus Added some test, and some error handling. Also removed the need to define core-templates in the conf (next will need to see how to extend this dinamically for dev).

PR #426, opened 2017-03-31
[GPT-565] Dynamic support to oc-templates in oc dev

nickbalestra:


src/utils/require-template.js
```diff
@@ -1,12 +1,21 @@
...
 var templateNotSupported = 'Error requiring the template "{0}": oc-template not found';
 
-module.exports = function(type) {
-  try {
-    return require(type);
-  } catch (err) {
-    throw format(templateNotSupported, type);
+module.exports = function(template) {  
+  var localTemplate = path.join(__dirname, '../../', 'node_modules', template);
+  if (fs.existsSync(localTemplate)) {
```
matteofigus: same as before. Can't use blocking code in the registry

src/cli/domain/get-components-deps.js
```diff
@@ -5,28 +5,41 @@ var path = require('path');
...
+    'handlebars': true
+  };
 
+  components.forEach(function(c){
     var pkg = fs.readJsonSync(path.join(c, 'package.json'));
+    var type = pkg.oc.files.template.type;
+    var dependencies = pkg.dependencies;
 
-    _.forEach(_.keys(pkg.dependencies), function(d){
+    if (!deps.templates[type] && !legacyTemplates[type]) {
```
nickbalestra: very nit and very arguable weather an array of 2 elements is more readable then an object with 2 keys.. I've removed the usage of `_.include` when reformatting this, we were doing way too many useless array loops, so my rationale is quicklookups are better

matteofigus: That's why I put it as question. I don't have strong opinions on this. Let's leave it as is if you like.

client/src/utils/require-template.js
```diff
@@ -1,12 +1,39 @@
...
+  // if (fs.existsSync(localTemplate)) {
+  //   return require(localTemplate);
+  // }
+
+  // var relativeTemplate = path.resolve('.', 'node_modules', template);
+  // if (fs.existsSync(relativeTemplate)) {
+  //   return require(relativeTemplate);
+  // }
+
+  // throw format(templateNotSupported, template);
```
matteofigus: forgot to remove this lines?

nickbalestra: 😅

nickbalestra:
@matteofigus should be ok, let me know if you see anything else that need moar 💅

src/registry/domain/repository.js
```diff
@@ -26,12 +27,7 @@ module.exports = function(conf){
   var coreTemplates = ['oc-template-jade', 'oc-template-handlebars'];
   var templates = _.union(coreTemplates, conf.templates)
     .map(function(template){
-      var info;
-      try {
-        info = require(template).getInfo();
-      } catch (err) {
-        throw new Error(format(strings.errors.registry.TEMPLATE_NOT_FOUND, template));
-      }
```
matteofigus: Let's re-add a try catch here for a template that, in case is not valid, may throw as getInfo doesn't exist or is not a function. Error wouldn't be "template not found" anymore but more of a "template not valid" kind?

src/cli/domain/get-components-deps.js
nickbalestra: Add check that if templatetype is not legacy must be declared as dependency

client/src/utils/require-template.js
```diff
@@ -1,12 +1,56 @@
 'use strict';
 
 var format = require('stringformat');
-var templateNotSupported = 'Error loading component: template "{0}" not supported';
+var path = require('path');
+var _ = require('underscore');
```
matteofigus: we don't have underscore in the client! Made the same mistake when reviewing the previous PR with @mattiaerre and then realised - are you able to go vanilla with client's code?

nickbalestra: Done, actually I like it even more without it.

PR #428, opened 2017-03-31
registerTemplate API fix

nickbalestra:
Found an interdependency issue that wasn't allowing pre registration of templates via config to work correctly.

`oc.registerTemplates` was not defined in the case of oc.conf.templates already avail.

### DIFF
--- a/src/components/oc-client/src/oc-client.js
+++ b/src/components/oc-client/src/oc-client.js
@@ -70,7 +70,7 @@ var oc = oc || {};
     }
   };
 
-  var coreTemplates = {
+  var registeredTemplates = {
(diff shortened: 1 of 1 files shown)


matteofigus:
given it's a bugfix, can you add a test to prove the bug is fixed?

nickbalestra:
Agree, testing oc [custom settings](https://github.com/opentable/oc/wiki/Browser-client#oc-custom-settings) like

```
<script>var oc = { conf: { retryInterval: 2000 }};</script>
```

 is not trivial though. It also seems to me that we don't have tests for similar config API, so perhaps is worth opening an issue to address this separately in another PR? Or am I missing something?

PR #429, opened 2017-03-31
[GPT-518] dynamically register templates on browser client

nickbalestra:
Now that we have registry supported [templates exposed in context](https://github.com/opentable/oc/pull/423). We could imagine to use this to programmatically and dynamically configure the browser client to support them [via the registerTemplate API](https://github.com/opentable/oc/pull/396).

- [x] extend browser oc-client component to use registerTemplate APIs.
- [x] dependency on #428
- [x] dependency on #426

### DIFF
--- a/src/components/oc-client/template.jade
+++ b/src/components/oc-client/template.jade
@@ -1,15 +1,3 @@
 script(src=staticPath+"src/oc-client.min.js", type="text/javascript")
 script.
-  oc.registerTemplates({
-    type: 'oc-template-react',
-    externals: [
-      {
-        "global": "React",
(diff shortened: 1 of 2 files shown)


src/components/oc-client/template.jade
```diff
@@ -1 +1,3 @@
-script(src=staticPath+"src/oc-client.min.js", type="text/javascript")
\ No newline at end of file
+script(src=staticPath+"src/oc-client.min.js", type="text/javascript")
+script.
+  oc.registerTemplates(!{JSON.stringify(templates)});
```
matteofigus: move to pre-config api with the usual client-side initialisation snippet

Issue #431, opened 2017-04-03
Component parameters are ignored when doing backend-rendering with oc-client nodejs library


gtrias:
I'm trying to send parameters to component in order to perform the backend rendering.  Given the next component array:

```
[{"name":"some-component","parameters":{"apikey":"1234abcd"}}]
```
The component is rendered but not receiving the parameter apikey. By the other hand if I request the component using curl and passing the parameter as a querystring the widget is taking it properly.
Am I missing anything?

oc-client version: ^2.1.6
oc registry version: 0.36.15

My current implementation is the next:

Component `server.js`:

```javascript
module.exports.data = function (context, callback) {
  callback(null, {
    staticPath: context.staticPath,
    params: context.params
  });
};
```
`template.html`
```html
{{params.apikey}}
```

Server side rendering:
```javascript
var Client = require('oc-client');

var client = new Client({
    registries: {
        serverRendering: 'http://myregistry.server',
        clientRendering: 'http://myregistry.public.server'
    }
});

var renderComponents = function (components, callback) {
  client.renderComponents(components, {}, function(err, html){
    if (err) {
      return callback(err);
    }

    callback(null, html);
  });
};
```

matteofigus:
What's the output of html? Can you console.log that and see what contains?

gtrias:
Given the next template:
```html
<div id="widget-selector"></div>
<script src="{{staticPath}}build/static/js/main.js"></script>
<script>
  // Instantiating widget, passing the selector where
  // you want to render it, and som parameters to configure it
  Instance = SomeWidget.default.widget.new({
    selector: "#some-selector",
    apikey: '{{params.apikey}}'
  });
  Instance.render();
</script>
```

The otuput of server-rendering is (changing sensitive values):

```html
<div id="widget-selector"></div>
<script src="//s3-eu-west-1.amazonaws.com/xxxxxxxx/components/xxxx-xxxxx/0.5.0/build/static/js/main.js"></script>
<script>
  // Instantiating widget, passing the selector where
  // you want to render it, and some parameters to configure it
  Instance = SomeComponent.default.widget.new({
    selector: "#widget-selector",
    apikey: ''
  });
  Instance.render();
</script>
<script>window.oc=window.oc||{};oc.renderedComponents=oc.renderedComponents||{};oc.renderedComponents["fees-admin"]="0.5.0";</script>
``` 
And using curl with querystring:
```html
[...]

PR #432, opened 2017-04-03
prepare-server-get-component-parameters

mattiaerre:
# Description

- add mergeObjects function to build the query string based on component and options parameters

- fix tests

- add `test:unit` as npm script


mattiaerre:
this PR should fix https://github.com/opentable/oc/issues/431 please @gtrias can you have a look at is and see if this makes sense to you?

// cc @matteofigus  @ivan-dimitrov-skyscanner

mattiaerre:
@matteofigus I've added a new function that merges 2 objects together and tests that should cover the scenario you've described; let me know what you think.

// @gtrias @ivan-dimitrov-skyscanner

gtrias:
@mattiaerre I'm going to test it right now. Thanks a lot for the quick fix ;)

gtrias:
Hey @mattiaerre how can I try your branch?
I'm trying to add `"oc-client": "git+ssh://git@github.com:opentable/oc.git#prepare-server-get-component-parameters"` and running `rm -rf node_modules/oc-client && npm install`.
But then it is installing all oc repository and nodejs is unable to find module `oc-client`:

```
Error: Cannot find module 'oc-client'
```
Is there a way to specify the client folder? AFAIK npm doesn't allow to do that https://github.com/npm/npm/issues/2974

client/src/href-builder.js
```diff
@@ -52,7 +53,11 @@ module.exports = function(config){
 
     prepareServerGet: function(baseUrl, component, options) {
       var urlPath = component.name + (component.version ? '/' + component.version : '');
-      var qs = options.parameters ? ('/?' + querystring.stringify(options.parameters)) : '';
+
+      var qs = '';
+      if (component.parameters || options.parameters) {
+        qs = '/?' + querystring.stringify(mergeObjects(component.parameters, options.parameters));
```
matteofigus: what about using _.extend. We use it a lot everywhere so it would be more consistent.
```js
var _ = require('underscore')
var p = _.extend(options.parameters, component.parameters)
// example
_.extend({b:3, c:4}, {a:1, b:2})
=> { b: 2, c: 4, a: 1 }
```

test/unit/client-href-builder.js
```diff
@@ -59,5 +59,16 @@ describe('client : href-builder :', () => {
             .to.equal('http://localhost:3030/hello-world/1.0.0/?p1=v1&p2=v%202');
       });
     });
+
+    describe('when there are parameters in both "options" and "component"', () => {
+      it('it should return a valid request for the component with the parameters set as URL query params', () => {
+        let options = { parameters: { p1: 'v1', p2: 'v 2' } };
+        let component = { name: 'hello-world', version: '1.0.0', parameters: { message: 'hello' } };
```
matteofigus: I would ensure we include a common parameter to ensure our merge goes in the right way, such as { message: 'hello', p1: 'v3'} and be sure qs has p1=v1

client/src/href-builder.js
mattiaerre: I was thinking about that but looking at the package.json I couldn't see `lodash` as a dependency (https://github.com/opentable/oc/blob/master/client/package.json#L17-L21) also in ES6 this will be part of the language so I preferred not to add a dependency and go "vanilla". plus the component is a standalone one (and fully tested) so we can always swap its internal implementation.

test/unit/client-href-builder.js
mattiaerre: this is covered at the `mergeObjects` level but I will add a specific test for this scenario.

mattiaerre:
over to you @matteofigus

gtrias:
@mattiaerre It works my use case now. But I haven't tried all the scenarios btw...

client/src/href-builder.js
matteofigus: we currently have underscore, not lodash

gtrias:
I can confirm using global parameters works as well but they can be overridden by the component ones...

matteofigus:
Awesome stuff thanks @mattiaerre

matteofigus:
Just published. You should be able to upgrade and test now @gtrias thanks for your help!

PR #433, opened 2017-04-03
[GPT-537]Cli init allows for --templateType to be npm module

nickbalestra:
If the templatetype passed to the init comand is not 'jade' or 'handlebars' (default) but a valid oc-template module, the cli will download it and use it to generate boilerplate.

[Example of boilerplate](https://github.com/opentable/oc-template-react/tree/wip/component) code used by the cli to generate a component

- [x] Coded to easily add yarn in the future 😍  (not yet supported)
- [x] Don't install template if already available on local dev registry
- [x] Use boilerplate code hosted template
- [x] add readme into boilerplates
- [x] logger dependcy injection
- [x] unit tests


nickbalestra:
Need to add some tests, but the CLI now allows to init from local modules or npm modules :)
🎉 

oc-dev

nickbalestra:
Created a [templates](https://github.com/opentable/oc/tree/templates) branch where all the work in relation to templates is going to be pushed

PR #443, opened 2017-04-06
Use baseUrl as default with host as fallback

matteofigus:
Fixes #413


debopamsengupta:
this looks great @matteofigus , fixes our problems 😄

PR #447, opened 2017-04-07
Components by dir cleanup

matteofigus:
See inline comments.

### DIFF
--- a/src/cli/domain/get-components-by-dir.js
+++ b/src/cli/domain/get-components-by-dir.js
@@ -11,13 +11,8 @@ module.exports = function(){
     var isOcComponent = function(file){
 
       var filePath = path.resolve(componentsDir, file),
-          packagePath = path.join(filePath, 'package.json');
-
-      if(!fs.existsSync(packagePath)){
-        return false;
-      }
-
-      var content;
+          packagePath = path.join(filePath, 'package.json'),
+          content;
 
       try {
         content = fs.readJsonSync(packagePath);
--- a/src/cli/domain/get-components-by-dir.js
+++ b/src/cli/domain/get-components-by-dir.js
@@ -4,47 +4,46 @@ var fs = require('fs-extra');
 var path = require('path');
 var _ = require('underscore');
 
-module.exports = function(dependencies){
-  var logger = dependencies.logger;
+module.exports = function(){
+
   return function(componentsDir, callback){
 
+    var isOcComponent = function(file){
+
+      var filePath = path.resolve(componentsDir, file),
+          packagePath = path.join(filePath, 'package.json');
+
+      if(!fs.existsSync(packagePath)){
+        return false;
+      }
+
+      var content;
+
+      try {
+        content = fs.readJsonSync(packagePath);
+      }
+      catch(err)
+      {
+        return false;
+      }
+
+      var packagedProperty = content.oc && content.oc.packaged;
+
+      return _.isUndefined(packagedProperty);
+    };
+
+    var dirContent;
+
     try {
-      var components = fs.readdirSync(componentsDir).filter(function(file){
-
-        var filePath = path.resolve(componentsDir, file),
-            isDir = fs.lstatSync(filePath).isDirectory(),
-            packagePath = path.join(filePath, 'package.json');
-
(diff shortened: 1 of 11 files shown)


src/cli/domain/get-components-by-dir.js
```diff
@@ -4,47 +4,46 @@ var fs = require('fs-extra');
 var path = require('path');
 var _ = require('underscore');
 
-module.exports = function(dependencies){
-  var logger = dependencies.logger;
-  return function(componentsDir, callback){
+module.exports = function(){
```
matteofigus: no reason logger should be accessible outside of cli facades.

src/cli/domain/get-components-by-dir.js
```diff
@@ -4,47 +4,46 @@ var fs = require('fs-extra');
...
 
-    try {
-      var components = fs.readdirSync(componentsDir).filter(function(file){
+  return function(componentsDir, callback){
 
-        var filePath = path.resolve(componentsDir, file),
-            isDir = fs.lstatSync(filePath).isDirectory(),
```
matteofigus: no reason to check <path> is a dir - let's just try to open <path>/package.json

src/cli/domain/get-components-by-dir.js
```diff
@@ -4,47 +4,46 @@ var fs = require('fs-extra');
...
+      return _.isUndefined(packagedProperty);
+    };
 
-      callback(null, fullPathComponents);
+    var dirContent;
 
+    try {
+      dirContent = fs.readdirSync(componentsDir);
     } catch(err){
-      return callback(err);
```
matteofigus: if the dir is not readable, let's skip it.

src/cli/facade/dev.js
```diff
@@ -140,10 +140,7 @@ module.exports = function(dependencies){
     log.warn(strings.messages.cli.SCANNING_COMPONENTS, true);
     local.getComponentsByDir(componentsDir, function(err, components){
 
-      if(err){
```
matteofigus: no way this is gonna error anymore

src/cli/domain/get-components-by-dir.js
```diff
@@ -4,47 +4,41 @@ var fs = require('fs-extra');
...
-        }
+      var filePath = path.resolve(componentsDir, file),
+          packagePath = path.join(filePath, 'package.json'),
+          content;
 
-        if(!content.oc || !!content.oc.packaged){
-          return false;
-        }
+      try {
+        content = fs.readJsonSync(packagePath);
```
matteofigus: Previous behaviour was to check dir existed, check package.json existed, and then trying to read. For simplicity, let's just try to read the Json in that folder. If it's a component, it will be there. If it's not, let's skip the dir.

src/cli/domain/get-components-by-dir.js
```diff
@@ -4,47 +4,41 @@ var fs = require('fs-extra');
...
+    };
 
-      callback(null, fullPathComponents);
+    var dirContent;
 
+    try {
+      dirContent = fs.readdirSync(componentsDir);
     } catch(err){
-      return callback(err);
+      return callback(null, []);
```
matteofigus: if the dir can't be read because is protected or doesn't exist, let's just say we didn't find anything.

src/cli/domain/local.js
```diff
@@ -11,8 +11,7 @@ var packageComponents = require('./package-components');
 var mock = require('./mock');
 var validator = require('../../registry/domain/validators');
 
-module.exports = function(dependencies){
-  var logger = dependencies.logger;
+module.exports = function(){
```
matteofigus: no way logger should be exposed outside of CLI facades...

src/cli/facade/dev.js
```diff
@@ -140,10 +140,7 @@ module.exports = function(dependencies){
...
-        callback(err);
-        return log.err(err);
-      } else if(_.isEmpty(components)){
+      if(_.isEmpty(components)){
```
matteofigus: This won't error anymore. In case no components are found, it will yield `(null, [])`

PR #448, opened 2017-04-07
[WIP][MLP-594] Adding history of published components to registry

federicomaffei:
Do not merge.

### DIFF
--- a/src/utils/history-sort.js
+++ b/src/utils/history-sort.js
@@ -1,0 +1,22 @@
+'use strict';
+
+var _ = require('underscore');
+var dateStringified = require('./date-stringify');
+
+module.exports = function(history) {
+  var result = [];
+  
+  _.each(history.components, function(singleComponent, name) {
+    _.each(singleComponent, function(release) {
+      var entry = {
+        name,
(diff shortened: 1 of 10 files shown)


mattiaerre:
@federicomaffei I like this feature

src/utils/history-sort.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+var _ = require('underscore');
+var dateStringified = require('./date-stringify');
+
+module.exports = function(history) {
+  var result = [];
+  
+  _.each(history.components, function(singleComponent, name) {
```
nickbalestra: I guess histories could grows quite long in time.. then for loops instead of nested each mill make quite a difference, we could also probably create a sorted array as we push them instead of having to loop through it again with sortBy. What you think?

federicomaffei: Agreed @nickbalestra, especially about the quick sort. That sort and reverse will not make it to the final version 👍

matteofigus:
Closed this in favour of #467 (the first commits came from this branch)

PR #449, opened 2017-04-07
No unused variables

matteofigus:
Removed some 💩 by setting no unused var setting in jshint and then correcting all the stuff.


src/cli/domain/get-missing-deps.js
```diff
@@ -22,7 +22,7 @@ module.exports = function(dependencies, components){
         delete require.cache[pathToModule];
       }
 
-      var required = require(pathToModule);
+      require(pathToModule);
```
mattiaerre: I do not understand what this piece of code does 😊

src/registry/index.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 var async = require('async');
-var colors = require('colors');
+var colors = require('colors/safe');
```
mattiaerre: new version of `colors`?

tasks/support/.jshintrc
```diff
@@ -17,7 +17,7 @@
     "plusplus"      : false,    // true: Prohibit use of `++` & `--`
     "quotmark"      : "single",     // Quotation mark consistency:
     "undef"         : true,     // true: Require all non-global variables to be declared (prevents global leaks)
-    "unused"        : false,    // true: Require all defined variables be used
+    "unused"        : true,    // true: Require all defined variables be used
```
mattiaerre: me ❤️  it

src/registry/index.js
matteofigus: colors exposes the classic mode, which just overrides global strings by allowing to do `"string".red` or the "safe" mode `colors.red("string")`. We use the safe mode almost everywhere but we missed this before..

src/cli/domain/get-missing-deps.js
matteofigus: The easiest way to know if you can require something is to wrap a require into a try/catch. Given the only value of this is to ensure this doesn't throw, no need to assign the value to a "required" variable (which, in fact, the linter finds as unused)

mattiaerre:
@matteofigus thanks, shall I merge this?

Issue #436, opened 2017-04-04
var => let/const


matteofigus:
A non critical optimisation that would be ideal for a first time contribution to OC.

Being very careful and perhaps splitting the work into multiple PRs to improve safety :rofl:

elboletaire:
Eslint may help with that:

- http://eslint.org/docs/rules/no-var
- http://eslint.org/docs/rules/prefer-const

matteofigus:
Yeah, actually @elboletaire you remind me that as part of this, we should probably move from jshint to eslint as it's much better for that!

nickbalestra:
=> https://github.com/prettier/prettier

elboletaire:
@matteofigus moving to eslint would be a good move, since jshint seems a bit outdated for all the ES6 stuff. Adding eslint to the project is as easy as `eslint --init`, and specify the source folders after asking some questions.

In fact I've done it locally but, there's something that scares me a bit... all that double negations found in [a][1] [lot][2] [of][3] [files][4]. Are they really needed? If they're there for casting to bools, wouldn't it be better to use ES5's `ToBoolean`?

Obviously I can ignore the warnings just changing a rule, although I'm still curious about it.

PS. If you want I can do a PR with the eslint added respecting the current "rules", and then another to change all the `var` with `const` and `let`.

[1]: https://github.com/opentable/oc/blob/master/src/cli/domain/watch.js#L12
[2]: https://github.com/opentable/oc/blob/f9f2dbb42329130237f2cceba554221836eb5751/src/cli/domain/get-missing-deps.js#L21
[3]: https://github.com/opentable/oc/blob/f9f2dbb42329130237f2cceba554221836eb5751/src/cli/domain/npm-installer.js#L8
[4]: https://github.com/opentable/oc/blob/master/src/cli/facade/dev.js#L25

matteofigus:
@elboletaire the !!castings definitely need to be checked one by one, it may be an hard work. I think most of them are not necessary.

I would love a PR for eslint, but I would first like to ensure everybody wants it. @nickbalestra you linked prettier which seems a nice project but I don't know so much about it. Do you think switching to eslint is something we still want to do it independently from prettier, or are the two things mutually exclusive?

mattiaerre:
@matteofigus I am not a prettier expert either but I think that if we incrementally improve our coding style, an eslint first step will not stop us to move towards prettier later if we want to. I, on the other hand, use eslint a lot and I see all the benefits.

[...]

PR #450, opened 2017-04-07
Replace jshint by eslint

elboletaire:
I'm creating the PR as a WIP because I'm currently working on it and I see things which create some questions for me:

- I get a lot of `no-undef` errors due to the injected libraries (under testing, the `it` and `describe`, more specifically). I've currently switch the rule to show warnings instead of errors, but I think it's better to disable it adding `/* eslint no-undef: 'off' */` for each test file that brings an error. If you're ok with it, I'll do a commit re-enabling the rule and setting the rule file by file.
- I've set `no-console` to `warn`. Would you like to set it to off, or is it ok to stay as warnings?

I'll also try to edit this post to add some notes about some decisions I made:

- The first commit is just the automatically created file from `eslint --init`. There are a lot of rules set to `off` that you may want to check.
- I've set `no-extra-boolean-cast` to `warn` in order to change from errors to warnings, so you don't forget about it :p
- I've set `no-empty` with` "allowEmptyCatch": true`, to allow empty `catch` statements.
- As there's no option for setting the `.eslintignore` file as a [`grunt-eslint`][grunt-eslint] parameter, I've decided to move both `.eslintrc.json` and `.eslintignore` to the root path, and remove the `tasks/support` folder, which was containing the `.jshintrc` file.
- We should consider changing the [`strict`][strict] rules, as they're not needed for nodejs modules (they're in strict mode by default).

refs #436

[grunt-eslint]: https://github.com/sindresorhus/grunt-eslint
[strict]: http://eslint.org/docs/rules/strict


matteofigus:
I've read your mind before so opened https://github.com/opentable/oc/pull/449 💃

elboletaire:
haha did not see it :p

Then I'll merge my branch with master when your PR is accepted. I remove that question from the PR, but please, try to answer the other ones too 😃

elboletaire:
Added note:

> - As there's no option for setting the `.eslintignore` file as a [`grunt-eslint`][grunt-eslint] parameter, I've decided to move both `.eslintrc.json` and `.eslintignore` to the root path, and remove the `tasks/support` folder, which was containing the `.jshintrc` file.

[grunt-eslint]: https://github.com/sindresorhus/grunt-eslint

elboletaire:
I've removed the `WIP` because my work is done until you answer that two questions in the PR 😃 

Tests run perfect locally, but for some reason have not been executed in travis nor in appveyor.

elboletaire:
PS. Merging with master, as #449 has been merged

elboletaire:
Added:

> - We should consider changing the [`strict`][strict] rules, as they're not needed for nodejs modules (they're in strict mode by default).

[strict]: http://eslint.org/docs/rules/strict

elboletaire:
Merged with master and fixed `no-unused-vars` issues appeared after enabling the rule again.

Note that I've set `no-unused-vars: 'off'` in [`registry-domain-components-cache.js`][registry-domain-components-cache.js] because there's giving a false positive (apparently does not properly detect the scope of the `response` var).

[registry-domain-components-cache.js]: https://github.com/elboletaire/oc/blob/ef81c2422bd1da17c1489ba8391b0ef7af86b89b/test/unit/registry-domain-components-cache.js#L2

matteofigus:
This looks good to me for now. Let's create separate issues for each rule that we may want to change. Agree about the no-strict, in case let's make a different PR for that.

Thanks for working on this!

PR #455, opened 2017-04-08
UI interface tests

matteofigus:


test/acceptance/registry-ui.js
```diff
@@ -0,0 +1,74 @@
...
+    result = r;
+    headers = d.response.headers;
+    done();
+  };
+
+  const conf = {
+    local: true,
+    path: path.resolve('test/fixtures/components'),
+    port: 3030,
+    baseUrl: 'http://localhost:3030/',
```
pbazydlo: would you be able to test it with a path, ex. `http://localhost:3030/v2/`?

matteofigus: I would be able to, not sure why should I?

The purpose of this tests is mostly ensure that we have a kind of smoke test for when we change the registry so that the UI doesn't break. Tests for testing the "prefix" parameter (what would be like a `/v2/` segment - are already there on a unit level.

pbazydlo: I've recently encountered a bug where `/v2/` was being removed when browsing components. But if it's covered elsewhere then it doesn't matter

matteofigus: Ok @pbazydlo, so can you approve?

PR #456, opened 2017-04-09
Removed trailing spaces

elboletaire:
Set eslint `no-trailing-spaces` to `error` and removed all trailing spaces.


elboletaire:
Ok now I see in the contribution guidelines that I should not merge my own PR... sorry.

BTW, as you can see this is a really minor change that will not affect the code at all 😇

matteofigus:
@elboletaire that is correct, I setup github permissions so that review will be required in the future, so this shouldn't happen anymore.

PR #459, opened 2017-04-11
Fix manual build script

matteofigus:
New acceptance tests need the version to be consistent in order for the UI to start without errors. It follows that build script needs to run before all the tests to run. This is also cleaner.


ArnoldZokas:
LGTM

PR #461, opened 2017-04-11
Registry start log fix

matteofigus:
A flip between colors/format in a previous PR made this to happen:
`Registry serving {0} components for a total of {1} releases. 157 3284`

This is to fix the message.


ArnoldZokas:
LGTM

PR #467, opened 2017-04-19
[MLP-594] History

matteofigus:
This allows us to have an history of all the published components. In the UI, there is a new tab that allows to navigate the history, with all the components sorted by date.

A couple of details about how it works:
* The complete history is saved in a new file hosted in the S3 bucket called `components-details.json`. In the future it may host much more than the publishedDate if necessary. That is a convenience view of the S3 components files tree.
* When the registry starts and creates the components list cache, it verifies the integrity of the components details (by comparing the two) and takes care of, in case, building the up-to-date components-details file
* The components details are not cached but just fetched when necessary
* When a new component is published, both components list and details files will be updated.
* The new history page can be quite huge in case of thousands of components and can benefit of a pagination component, not included in this PR.

For the review, I don't recommend going through all the commits - look at the diff and inline comments.

Home Page with 2 tabs:
screen shot 2017-04-26 at 13 59 29

History page:
screen shot 2017-04-26 at 13 59 46


src/registry/domain/components-details.js
```diff
@@ -0,0 +1,73 @@
...
+      }, done);
+    }, (err) => callback(err, {
+      lastEdit: getUnixUTCTimestamp(),
+      components: details.components
+    }));
+  };
+
+  const save = (data, callback) => cdn.putFileContent(JSON.stringify(data), filePath(), true, callback);
+
+  const refresh = (componentsList, callback) => {
```
matteofigus: The refresh signature (called on registry start and component published) has the components list (all the components and all versions) as input. The reason is that both times this is needed to happen, it happens after the list is fetched or modified, so it is good to start from there for navigating the tree.

src/registry/domain/repository.js
```diff
@@ -183,7 +189,10 @@ module.exports = function(conf){
         return callback(null, 'ok');
       }
 
-      componentsCache.load(callback);
+      componentsCache.load((err, componentsList) => {
+        if(err){ return callback(err); }
+        componentsDetails.refresh(componentsList, (err) => callback(err, componentsList));
+      });
```
matteofigus: This is what happens on registry start. In this scenario, the `componentsDetails.refresh` main goal is to verify integrity or to build the file from scratch (only the first time).

src/registry/domain/repository.js
```diff
@@ -230,12 +239,12 @@ module.exports = function(conf){
 
         cdn.putDir(pkgDetails.outputFolder, `${conf.s3.componentsDir}/${componentName}/${componentVersion}`, (err) => {
           if(err){ return callback(err); }
-          componentsCache.refresh(callback);
+          componentsCache.refresh((err, componentsList) => {
+            if(err){ return callback(err); }
+            componentsDetails.refresh(componentsList, callback);
+          });
```
matteofigus: On component published, both files are in need to be updated and succeed otherwise the publish will fail and the CLI will error

src/registry/domain/repository.js
```diff
@@ -230,12 +239,12 @@ module.exports = function(conf){
...
         });
       });
-    },
-    saveComponentsInfo: (componentsInfo, callback) => {
-      cdn.putFileContent(JSON.stringify(componentsInfo), `${conf.s3.componentsDir}/components.json`, true, callback);
```
matteofigus: this method had 0 usages so it was safe to be removed

src/registry/router.js
```diff
@@ -7,7 +7,7 @@ const ComponentRoute = require('./routes/component');
 const ComponentsRoute = require('./routes/components');
 const ComponentInfoRoute = require('./routes/component-info');
 const ComponentPreviewRoute = require('./routes/component-preview');
-const ListComponentsRoute = require('./routes/list-components');
+const IndexRoute = require('./routes/index');
```
matteofigus: The UI's "home page" used to be called "list components" route but now it's becoming much more than that. So list components => index

src/registry/routes/helpers/get-components-history.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
```
matteofigus: A little utility that given the details file maps to a convenient sorted array.

src/registry/views/index.jade
```diff
@@ -0,0 +1,30 @@
...
+  +property('Registry type', isLocal ? 'Local dev registry' : 'On-line registry')
+
+  h2#menuList
+    a(href="#components-list" class="tab-link") Components
+    if componentsHistory
+      | &nbsp;|&nbsp;
+      a(href="#components-history" class="tab-link") History
+  
+  include components-list
+  include components-history
```
matteofigus: There are 2 jade partials, and via client-side javascript we toggle the selected one

test/unit/registry-domain-events-handler.js
```diff
@@ -3,7 +3,7 @@
 const expect = require('chai').expect;
 const sinon = require('sinon');
 
-describe('registry : events-handler', () => {
+describe('registry : domain : events-handler', () => {
```
matteofigus: unrelated, consistency renaming

src/registry/domain/components-details.js
```diff
@@ -0,0 +1,73 @@
...
+    return callback(code);
+  };
+
+  const filePath = () => `${conf.s3.componentsDir}/components-details.json`;
+
+  const getFromJson = callback => cdn.getJson(filePath(), true, callback);
+
+  const getFromDirectories = (options, callback) => {
+
+    const details = _.extend({}, _.cloneDeep(options.details));
```
nickbalestra: Any reason why extending + deep cloning? (`cloneDeep` already return a new object if i'm not wrong)

matteofigus: options.details can be empty in case it's the first time the new version is running with the current registry. So, options.details can be undefined and I need details to fallback to empty object in that scenario.

Issue #470, opened 2017-04-23
issue w/ component info get params


mattiaerre:
# Description

v0.36.26 introduced a bug at the registry level for a component's preview. In particular the method `getParams` of the `component-info` route, always returns an empty object; this is preventing the `~info` route to preview a component w/ mandatory parameters.

v0.36.25
<img width="845" alt="screen shot 2017-04-22 at 18 02 27" src="https://cloud.githubusercontent.com/assets/6423261/25309585/ff5a44d0-2785-11e7-8b02-27d481bd9f72.png">

v0.36.26
<img width="848" alt="screen shot 2017-04-22 at 18 03 10" src="https://cloud.githubusercontent.com/assets/6423261/25309586/03e1f0c0-2786-11e7-83f6-a8964acdaac5.png">

I will create a PR w/ the fix

// cc @matteofigus @nickbalestra

mattiaerre:
I believe this issue has been introduced by this line of code:

https://github.com/opentable/oc/commit/5e942aacfbe10d0c205e9447ce9cf520be71e485#diff-c7c711fc0e33358b368c4394b68a3914

PR #471, opened 2017-04-23
[WIP] issue-with-component-info-get-params

mattiaerre:
# Description

as per #470 this PR adds a helper that can be used by `/src/registry/routes/component-info.js` in order to return the component's mandatory parameters as a map.

the integration hasn't been done yet as this PR only highlights the existing bug and compares:

1. `getParams` w/ `lodash` (aka `getParams`)

2. `getParams` w/ `underscode` (aka `getParams__`)

3. `getParams` w/o any dependencies (aka `getParamsV2`)

where

**1** has a bug, **2** and **3** behave the same

## Tests

screen shot 2017-04-22 at 18 05 34

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "stringformat": "0.0.5",
     "targz": "1.0.1",
     "uglify-js": "2.6.4",
+    "underscore": "^1.8.3",
     "watch": "0.19.1",
     "webpack": "2.2.0",
     "yargs": "^6.6.0"
--- a/src/registry/routes/helpers/get-params.js
+++ b/src/registry/routes/helpers/get-params.js
@@ -29,7 +29,7 @@ function getParams__(component) {
   return params;
 }
 
-const getParamsV2 = (component) => {
+function getParamsV2(component) {
   const params = {};
   if (component.oc.parameters) {
     Object.keys(component.oc.parameters).forEach((key) => {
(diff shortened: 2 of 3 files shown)


mattiaerre:
So it looks like that also `lodash` has got a `pick` function; so we can fix the issue using `pick` instead of `pickBy` here:

https://github.com/opentable/oc/blob/master/src/registry/routes/component-info.js#L18

mattiaerre:
as per my last commit (finally) the 3 functions are producing the same result; I would prefer to use the `getParamsV2` version as it doens't require any additional dependencies but if you think that using `pick` in the existing component is better, I am also totally fine w/ that.

matteofigus:
Hi, thanks for taking care of this.

1) Travis is complaining as you updated package but not the shrinkwrap 👍 
2) I think my preference would be to use lodash's `pick` as that's what we decided when doing the _ migration. This was erroneously converted to pickBy.
3) In either case we didn't catch this because of a lack of tests on that area, so I think it would be nice to keep your test together with the fix.
4) I would like to prevent issues like this to happen in the future and I believe improving test coverage would help. Do you think we can open a new issue to discuss test coverage and perhaps evaluate some tools to track the progress with that?

mattiaerre:
Hi @matteofigus sure, I 100% agree, let's keep it simple. I am going to close this PR and open a new one w/ just the fix. Re testing this part is quite tricky w/o moving the function to a separate file and I am not sure if this would make any difference. I like the idea of an issue w/ some sort of discussion for a better testing strategy; you know how much I ❤️ this subject 😄

mattiaerre:
Closing because of #472

Issue #476, opened 2017-04-27
get-components-by-dir issue


mattiaerre:
# Description

after the release v0.36.16, if you try to run `oc dev .` where there is a folder w/ a package.json file in it and this folder doesn't contain a component, you get the following error:

```bash
Cannot read property 'files' of undefined
```

this is because the function `isOcComponent` returns true also if the `oc` section is missing in the package.json.

I believe that the problem has been introduced w/ this PR:

[https://github.com/opentable/oc/pull/447](https://github.com/opentable/oc/pull/447)

and in particular:

[https://github.com/opentable/oc/pull/447/files#diff-6810304ebc65ef9905ca3d7d77d8a311](https://github.com/opentable/oc/pull/447/files#diff-6810304ebc65ef9905ca3d7d77d8a311)

mattiaerre:
I am going to create a PR for this

PR #477, opened 2017-04-27
get-components-by-dir-issue

mattiaerre:
# Description

as per #476, this PR checks if there is the `oc` section in the package.json and returns false if missing.


matteofigus:
Thanks @mattiaerre

PR #478, opened 2017-05-04
Upgrade chai

matteofigus:


test/acceptance/client.js
```diff
@@ -600,7 +600,6 @@ describe('The node.js OC client', () => {
         });
 
         it('should return non rendered contents', () => {
-          expect($component).to.exist();
```
matteofigus: exist has been deprecated. In any case, the second test is enough to ensure the $component element is not empty.

nickbalestra: Good catch

nickbalestra:
LGTM 👍

PR #479, opened 2017-05-04
Allow node core deps to be used and whitelisted

### LINKED ISSUE #462: Allow node's url core dependency on server.js
Requested by @jankowiakmaria 

I think url is stable and safe to be whitelisted.

matteofigus:
Fixes #462

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "babili-webpack-plugin": "0.0.11",
     "basic-auth-connect": "^1.0.0",
     "body-parser": "^1.16.0",
+    "builtin-modules": "^1.1.1",
     "clean-css": "3.4.18",
     "colors": "1.1.2",
     "dependency-graph": "0.5.0",
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "semver-extra": "2.0.1",
     "stringformat": "0.0.5",
     "targz": "1.0.1",
+    "try-require": "^1.2.1",
     "uglify-js": "2.6.4",
     "watch": "0.19.1",
     "webpack": "2.2.0",
--- a/src/cli/domain/get-components-deps.js
+++ b/src/cli/domain/get-components-deps.js
@@ -1,9 +1,11 @@
 'use strict';
 
+const coreModules = require('builtin-modules');
+const format = require('stringformat');
 const fs = require('fs-extra');
 const path = require('path');
 const _ =  require('lodash');
-const format = require('stringformat');
+
 const settings = require('../../resources');
 
 module.exports = function(components){
@@ -28,9 +30,7 @@ module.exports = function(components){
 
     _.keys(dependencies).forEach((name) => {
       const version = dependencies[name];
-      const depToInstall = version.length > 0
-        ? (name + '@' + version)
-        : name;
+      const depToInstall = version.length > 0 ? `${name}@${version}` : name;
 
       if (!deps.withVersions[depToInstall]) {
         deps.withVersions[depToInstall] = true;
@@ -43,7 +43,7 @@ module.exports = function(components){
   });
 
   return {
-    modules: _.keys(deps.modules),
+    modules: _.union(coreModules, _.keys(deps.modules)),
     withVersions: _.keys(deps.withVersions),
     templates: _.keys(deps.templates)
   };
--- a/src/cli/domain/get-components-deps.js
+++ b/src/cli/domain/get-components-deps.js
@@ -19,7 +19,7 @@ module.exports = function(components){
   components.forEach((c) => {
     const pkg = fs.readJsonSync(path.join(c, 'package.json'));
     const type = pkg.oc.files.template.type;
-    const dependencies = pkg.dependencies;
+    const dependencies = pkg.dependencies || {};
 
     if (!deps.templates[type] && !legacyTemplates[type]) {
       if (!dependencies[type]) {
@@ -43,7 +43,7 @@ module.exports = function(components){
   });
 
   return {
-    modules: _.union(coreModules, _.keys(deps.modules)),
+    modules: _.union(coreModules, _.keys(deps.modules)).sort(),
     withVersions: _.keys(deps.withVersions),
     templates: _.keys(deps.templates)
   };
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "babili-webpack-plugin": "0.0.11",
     "basic-auth-connect": "^1.0.0",
     "body-parser": "^1.16.0",
+    "builtin-modules": "^1.1.1",
```
matteofigus: a little convenience module with a list of all the core node dependencies

src/cli/domain/get-components-deps.js
```diff
@@ -17,7 +19,7 @@ module.exports = function(components){
   components.forEach((c) => {
     const pkg = fs.readJsonSync(path.join(c, 'package.json'));
     const type = pkg.oc.files.template.type;
-    const dependencies = pkg.dependencies;
+    const dependencies = pkg.dependencies || {};
```
matteofigus: this comes from package.json. After testing, I found a component with no dependencies props could actually generate some troubles

src/cli/domain/get-components-deps.js
```diff
@@ -43,7 +43,7 @@ module.exports = function(components){
   });
 
   return {
-    modules: _.keys(deps.modules),
+    modules: _.union(coreModules, _.keys(deps.modules)).sort(),
```
matteofigus: This is used for the CLI. Now the cli by default allows any core module to be used. When running in the registry, still, the dependency will need to be whitelisted as all the other deps in order to be usable.

src/cli/domain/package-server-script/bundle/config/external-dependencies-handlers.js
```diff
@@ -5,18 +5,21 @@
...
 
 
 module.exports = function externalDependenciesHandlers(dependencies){
   const deps = dependencies || {};
 
-  const missingExternalDependecy = function(dep, dependencies) {
-    return !_.includes(_.keys(dependencies), dep);
-  };
+  const missingExternalDependency = (dep, dependencies) =>
+    !_.includes(_.keys(dependencies), dep) && !_.includes(coreModules, dep);
```
matteofigus: This part is used for the CLI bundler. This is for making the bundler bypass the dependency in case is a core, by being treated in the same way normal deps are.

src/cli/domain/package-server-script/bundle/config/index.js
```diff
@@ -3,7 +3,7 @@
 
 const webpack = require('webpack');
 const path = require('path');
-const externalDependenciesHandlers = require('./externalDependenciesHandlers');
+const externalDependenciesHandlers = require('./external-dependencies-handlers');
```
matteofigus: just a little renaming for consistency

src/registry/domain/require-wrapper.js
```diff
@@ -1,32 +1,44 @@
...
     try {
-      return require(modulePath);
+      return requireCoreDependency(requirePath);
     } catch (e) {
-      throw {
-        code: strings.errors.registry.DEPENDENCY_NOT_FOUND_CODE,
-        missing: [modulePath]
-      };
+      throw getError(requirePath);
     }
```
matteofigus: So the flow here is 1) try to require the module normally 2) if it fails and is a core dep, try to require it 3) if it fails, return the error at the end

src/cli/domain/get-components-deps.js
nickbalestra: Why are we sorting them?

src/registry/domain/require-wrapper.js
```diff
@@ -1,32 +1,44 @@
...
+  if(_.includes(coreModules, requirePath)){
+    return require(requirePath);
+  }
+  throw new Error();
+};
+
+const requireDependency = (requirePath) => {
+  const nodeModulesPath = path.resolve('.', 'node_modules');
+  const modulePath = path.resolve(nodeModulesPath, requirePath);
+  return require(modulePath);
```
nickbalestra: I'll perhaps move the try catch logic inside each requiring function `requireCoreDependency` and `requireCoreDependency` .

src/registry/domain/require-wrapper.js
```diff
@@ -1,32 +1,44 @@
...
+  } catch (e) {
     try {
-      return require(modulePath);
+      return requireCoreDependency(requirePath);
     } catch (e) {
-      throw {
-        code: strings.errors.registry.DEPENDENCY_NOT_FOUND_CODE,
-        missing: [modulePath]
-      };
+      throw getError(requirePath);
```
nickbalestra: `getError` accept a param called moduleName and here you are passing a path, perhaps worth renaming such param to accomodate both cases better?

src/registry/domain/require-wrapper.js
nickbalestra: 👍

src/cli/domain/get-components-deps.js
matteofigus: Good point.

src/registry/domain/require-wrapper.js
matteofigus: Ok done. I found a `try-require` module for that and made some more cleanup around.

src/registry/domain/require-wrapper.js
matteofigus: Ok done

src/registry/domain/require-wrapper.js
nickbalestra: Nice we should then use it also for the templates as we are doing similar stuff in multiple places, perhaps once we work on allowing multiple-version deps

src/registry/domain/require-wrapper.js
```diff
@@ -2,43 +2,35 @@
...
-  try {
-    return requireDependency(requirePath);
-  } catch (e) {
-    try {
-      return requireCoreDependency(requirePath);
-    } catch (e) {
-      throw getError(requirePath);
-    }
-  }
+  return requireDependency(requirePath) || requireCoreDependency(requirePath) || throwError(requirePath);
```
nickbalestra: love it

PR #482, opened 2017-05-11
Upgrade multer

matteofigus:
We use a very old version of multer.

### DIFF
--- a/src/registry/domain/extract-package.js
+++ b/src/registry/domain/extract-package.js
@@ -2,15 +2,14 @@
 
 const path = require('path');
 const targz = require('targz');
-const _ = require('lodash');
 
 const getPackageJsonFromTempDir = require('./get-package-json-from-temp-dir');
 
 module.exports = function(files, callback){
 
-  const packageFile = files[_.keys(files)[0]],
+  const packageFile = files[0],
     packagePath = path.resolve(packageFile.path),
-    packageUntarOutput = path.resolve(packageFile.path, '..', packageFile.name.replace('.tar.gz', '')),
+    packageUntarOutput = path.resolve(packageFile.path, '..', packageFile.filename.replace('.tar.gz', '')),
     packageOutput = path.resolve(packageUntarOutput, '_package');
 
   targz.decompress({
--- a/src/registry/domain/validators/uploaded-package.js
+++ b/src/registry/domain/validators/uploaded-package.js
@@ -3,9 +3,7 @@
 const _ = require('lodash');
 
 module.exports = function(input){
-  const response = {
-    isValid: true
-  };
+  const response = { isValid: true };
 
   const returnError = function(message){
     response.isValid = false;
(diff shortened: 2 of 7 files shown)


src/registry/domain/extract-package.js
```diff
@@ -2,15 +2,14 @@
...
 const path = require('path');
 const targz = require('targz');
-const _ = require('lodash');
 
 const getPackageJsonFromTempDir = require('./get-package-json-from-temp-dir');
 
 module.exports = function(files, callback){
 
-  const packageFile = files[_.keys(files)[0]],
+  const packageFile = files[0],
```
matteofigus: instead of object, `req.files` now returns an array

src/registry/domain/validators/uploaded-package.js
```diff
@@ -21,9 +19,9 @@ module.exports = function(input){
     return returnError('not_valid');
   }
 
-  const file = input[_.keys(input)[0]];
+  const file = input[0];
 
-  if(file.mimetype !== 'application/octet-stream' || !!file.truncated || file.extension !== 'gz' || file.path.indexOf('.tar.gz') < 0){
```
matteofigus: `file.extension` is deprecated and `file.filename` already has the extension check

src/registry/middleware/file-uploads.js
```diff
@@ -9,11 +8,17 @@ module.exports = function(req, res, next){
     return next();
   }
 
-  return multer({
-    dest: res.conf.tempDir,
-    fieldSize: 10,
-    rename: function(fieldname, filename){
-      return format('{0}-{1}.tar', filename.replace('.tar', '').replace(/\W+/g, '-').toLowerCase(), Date.now());
```
matteofigus: The old filename replacement didn't take in consideration the extension (`.gz`).
The new mechanism has the whole filename so that's why it's slightly different.

test/unit/registry-domain-extract-package.js
```diff
@@ -27,12 +27,10 @@ describe('registry : domain : extract-package', () => {
...
       decompressStub.yields();
 
-      extractPackage({
-        '1478279453422.tar.gz': {
-          name: '1478279453422.tar.gz',
-          path: '/some-path/registry/temp/1478279453422.tar.gz'
-        }
-      }, (err, res) => {
+      extractPackage([{
+        filename: '1478279453422.tar.gz',
```
matteofigus: `name` => `filename`

src/registry/domain/extract-package.js
```diff
@@ -2,15 +2,14 @@
...
     packagePath = path.resolve(packageFile.path),
-    packageUntarOutput = path.resolve(packageFile.path, '..', packageFile.name.replace('.tar.gz', '')),
+    packageUntarOutput = path.resolve(packageFile.path, '..', packageFile.filename.replace('.tar.gz', '')),
```
matteofigus: `name` => `filename`

src/registry/domain/validators/uploaded-package.js
```diff
@@ -21,9 +19,9 @@ module.exports = function(input){
...
+  if(file.mimetype !== 'application/octet-stream' || file.truncated || file.filename.indexOf('.tar.gz') < 0){
```
mattiaerre: minor but for readability purpose what about start to use `contains` instead? I know we are still targeting node 4. But just as a reminder 😊 

matteofigus: Sure. Perhaps we can switch to lodash's contains in a separate PR using this: https://github.com/wix/eslint-plugin-lodash/blob/master/docs/rules/prefer-includes.md

And then move to ES6's when deprecating node 4?

nickbalestra: +1 on lodash's include

PR #484, opened 2017-05-11
[GPT-645] Handling data-provider compilation within template

nickbalestra:
As per discussed on https://github.com/opentable/oc/issues/439L:

> template expose a compileServer() method that will be used instead of package-server-script/bundle


nickbalestra:
Merged data-provider compilation in the UNIFIED compiler API, final work for this is being done there

PR #488, opened 2017-05-24
yarn-support

mattiaerre:
# Description

see #487

- get template by name

- fix test

I've added a **require by name** step before the existing **require by path** ones; I've also added some small refactoring. I've tested this fix locally and it looks like it works w/ both npm and yarn.

## Coverage report

```bash
 PASS  __tests__/src/utils/require-template.test.js
  requireTemplate
    ✓ expect type of `requireTemplate` to be function (3ms)
    valid
      ✓ oc-template-handlebars (68ms)
      ✓ oc-template-jade (267ms)
    not valid
      ✓ lodash (35ms)
      ✓ oc-invalid-template (3ms)

Test Suites: 1 passed, 1 total
Tests:       5 passed, 5 total
Snapshots:   4 passed, 4 total
Time:        1.432s
Ran all test suites.
---------------------|----------|----------|----------|----------|----------------|
File                 |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
---------------------|----------|----------|----------|----------|----------------|
All files            |       96 |     87.5 |      100 |       96 |                |
 require-template.js |       96 |     87.5 |      100 |       96 |             24 |
---------------------|----------|----------|----------|----------|----------------|
```


src/utils/require-template.js
```diff
@@ -19,30 +19,34 @@ function isValidTemplate(template){
...
   const localTemplate = path.join(__dirname, '../../', 'node_modules', template);
   const relativeTemplate = path.resolve('.', 'node_modules', template);
 
   try {
-    if (require.cache && !!require.cache[localTemplate]) {
-      delete require.cache[localTemplate];
-    }
-    ocTemplate = require(localTemplate);
-  } catch(err) {
+    ocTemplate = getOcTemplate(template);
```
mattiaerre: this is the only "new" code I've added that tries to fix the yarn issue.

elboletaire: Sorry about my ignorance, but how can that change affect over the yarn installation behavior? I only see you moved the logic to a function 😕

mattiaerre: Hi,
Yes I moved the logic into a function but I've also added an additional check of getting the template by name (line 35). 
Try to do this: install OC globally w/ yarn and make sure your publish doesn't work. Then replace require-template.js in the yarn OC global folder w/ the file I've changed and see if that fixes. Thanks!

elboletaire: I've just tried it, and I can't publish due to this: https://github.com/npm/npmlog/issues/48

imatge

😕

mattiaerre: this is another problem; you need to rm -rf node_modules first, it is a known issue w/ npmlog when switching between npm and yarn 😢

elboletaire: Yeah, I was doing it now... and then I face with this...

imatge

After seeing the error... I thought that maybe your changes now could allow me to publish from the widget folder itself, so I tried:

imatge


the neverending story... 😅

mattiaerre: `infinite-loop-loader` tells us everything hehehe 😊 feel free to collaborate on this branch; I will not be able to follow this issue as I'll be offline for a couple of days but either @matteofigus or @nickbalestra should be able to have a look at this.

elboletaire: After manually installing `infinite-loop-loader` (ofc only for this test) everything works as expected! 😃

mattiaerre: very good; we are getting there 😉 thanks for your help!

mattiaerre:
@matteofigus @nickbalestra @elboletaire not sure if we decided how to proceed w/ this issue; `infinite-loop-loader` was the last blocker we had.

mattiaerre:
@matteofigus shall we have a conversation and decide how to proceed w/ this?

// cc @nickbalestra @elboletaire

matteofigus:
@mattiaerre sorry for taking long to reply to this.

Unfortunately requiring by name doesn't work. You want to require the template from the CLI local to the component. Requiring by name requires it local to the CLI, which happens to work currently for the jade/handlebars ones, but that will change soon in favour of taking those two off from the CLI (making the require by name attempt basically not useful).

Now, getting back to the issue. I am not very expert about yarn, but this seems to be a yarn problem to me. There is a package.json, yarn is not installing all the deps during the install, so I would open a issue in their repo rather than here.

matteofigus:
@nickbalestra please can you review what I just said makes sense? lol

src/utils/require-template.js
```diff
@@ -17,7 +17,14 @@ function isValidTemplate(template) {
...
+const getOcTemplate = (path) => {
+  if (require.cache && !!require.cache[path]) {
+    delete require.cache[path];
+  }
+  return require(path);
+};
+
+module.exports = function (template) {
   let ocTemplate;
   const localTemplate = path.join(
```
matteofigus: I think we need to understand, in case of yarn install, why this doesn't work. The oc-template-jade module must be there if yarn installed everything correctly, if it is not we need to find out where it is

matteofigus: This module https://github.com/substack/node-resolve seems like could be good for the require template functionality, /cc @nickbalestra so that we could change both local and relative templates to look into those paths and find the dependency. This would allow us to have a consistent dev experience between different versions of npm (some flattening the tree, some not) and yarn?

matteofigus: Same for the infinite-loop-loader. It should be in the node_modules.

matteofigus:
Ok...I apology for previous comments. 
I finally managed to reproduce.
Summarising:
`yarn global add oc` installs oc into => `/Users/mfigus/.nvm/versions/node/v6.10.2/bin/oc`
That thing is a symlink to => `/Users/mfigus/.config/yarn/global/node_modules/oc/`

Now...If I go in that node_modules folder, it's pretty empty. The main reason, is that if I go 1 level down (`/Users/mfigus/.config/yarn/global/node_modules/`) the dependencies (including infinite-loop-loader and oc-template-handlebars) are there already. So, I believe yarn tries to be smart and doesn't install them, but in the module we actually require it from the oc's node_modules.

So, in conclusion, I believe @mattiaerre's approach in this PR is correct and will fix.
So, I am approving this PR.

elboletaire:
+1 to what @matteofigus said about the `yarn.lock` file. It's ok to include it in final projects, but not in libraries (exactly how `composer.lock` works for php projects). So it should not be included in oc.

And I also think that, if we don't currently have testing, we should not merge issues. Let's add a new issue for adding tests. Given the moment, @mattiaerre will be able to cherry-pick the changes from his branch to the new PR.

mattiaerre:
thank you for your feedback @elboletaire I adapted my tests to use mocha and it looks like I've been able to achieve the same result w/o Jest (although no snapshots, unfortunately); apart from that I think that this is a decent refactoring so as long as we are good w/ `infinite-loop-loader` this PR is over to you for a review and no longer WIP

a quote from @ericelliott

> Open source etiquette requires a "when in Rome" attitude. Don't make major changes to tech or style without discussing it first.

source: [twitter](https://twitter.com/_ericelliott/status/875534668078223361)

matteofigus:
Nice quote 👍 It now LGTM. If it's good for you too, @elboletaire, can you merge?

Thanks

elboletaire:
Done @matteofigus

mattiaerre:
@elboletaire thank you!

Issue #445, opened 2017-04-06
Plugins detective is broken


matteofigus:
I believe the plugin detective used to parse the minified ES5 server.js to try to catch how `context` was transformed, and then would seek for `(that).plugins.(something)` calls in order to detect calls to not supported plugins. Now that we are bundling and compiling with Babili the Regex doesn't work anymore: https://github.com/opentable/oc/blob/master/src/registry/domain/plugins-detective.js#L7

Maybe we can discuss how to improve that. If we can't, we can rather remove the check and let the registry throw (which would then fire a domain error with a component's 5xx response).

mattiaerre:
@matteofigus happy to have a look at this issue if you like

matteofigus:
ok! thanks

mattiaerre:
@matteofigus it looks like `detective` is only used here:

https://github.com/opentable/oc/blob/master/src/registry/routes/helpers/get-component.js#L341

inside the `handleError` function that is raised only if `domain.on('error', handleError)`:

https://github.com/opentable/oc/blob/master/src/registry/routes/helpers/get-component.js#L365

so the detective logic is only exercised if there is an error; if I do not register a plugin and I try to get a component that uses one, this `on error` is never raised because there is no error; I've got the feeling that the detective maybe still work but the `get-component` functionality doesn't behave as expected; or am I missing something?

mattiaerre:
@matteofigus as #495 has been merged shall we close this issue?

matteofigus:
Oh yeah. Thanks!

PR #495, opened 2017-05-30
issue w/ plugins detective

mattiaerre:
# Description

as per #445 this functionality is broken; this PR tries to fix the `parse` logic. This, unfortunately, will not fix the fact that the Registry still will not throw any exception for the missing plugin(s); have a look at the issue mentioned above for [more details](https://github.com/opentable/oc/issues/445#issuecomment-304744155).


mattiaerre:
@matteofigus ... 😊

test/unit/code.js
```diff
@@ -0,0 +1 @@
+module.exports = "module.exports=function(a){function b(d){if(c[d])return c[d].exports;var e=c[d]={i:d,l:!1,exports:{}};return a[d].call(e.exports,e,e.exports,b),e.l=!0,e.exports}var c={};return b.m=a,b.c=c,b.i=function(d){return d},b.d=function(d,e,f){b.o(d,e)||Object.defineProperty(d,e,{configurable:!1,enumerable:!0,get:f})},b.n=function(d){var e=d&&d.__esModule?function(){return d['default']}:function(){return d};return b.d(e,'a',e),e},b.o=function(d,e){return Object.prototype.hasOwnProperty.call(d,e)},b.p='',b(b.s=0)}([function(a,b){'use strict';Object.defineProperty(b,'__esModule',{value:!0});b.data=function(f,g){var h={logo:{href:'',url:f.params.logoUrl,alt:f.params.title+' logo'},title:f.params.title,timezoneOffset:'?'};f.plugins.getTimezoneOffset&&(h.timezoneOffset=f.plugins.getTimezoneOffset()),g(null,h)}}]);";
```
matteofigus: Can we move this file to `/test/fixtures`? I think that's supposed to be the place for this kind of files.

mattiaerre: sure, although I like to keep related things together 😊

matteofigus: I think my preference would be to make it a component and bring it to `/test/fixtures/mocked-components/missing-plugin.js` (look at the others in that dir to follow same convention)

mattiaerre: @matteofigus I will try to do that although what I do actually need here is only generated code, but I see your point

test/unit/registry-domain-plugins-detective.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const expect = require('chai').expect;
+const code = require('../fixtures/mocked-components/package-server-with-plugin');
```
mattiaerre: @matteofigus what do you think about this? I only added the `data` part and not the full object; you know that I'm a minimalist 😅

Issue #404, opened 2017-03-13
get rid of npm-shrinkwrap


matteofigus:
So, I am creating an issue to track this. 

#### why the shrinkwrap?
To be clear, the reason we have the shrinkwrap atm is to override some subdependencies on the jade module that are unsecure (old versions of uglify-js). This is bad because every module change needs a rework on the npm-shrinkwrap, and to limit bad scenarios and quick PR reviews we need to make surgical changes to that file in that regard.

#### future plan
get rid of it when #298 will be complete.

#### in the meanwhile
when working on a dependency,
* use npm 2 by `sudo npm i npm@2`
* run `npm install <module> --save`
* diff the npm-shrinkwrap file and rework it manually in order to avoid any changes to other modules, only the ones about the module added or upgraded
* check in the change in the PR

matteofigus:
Update: I found this awesome module: https://github.com/jnordberg/jade-legacy which is jade + security patches, all green on snyk: https://snyk.io/test/github/jnordberg/jade-legacy

Perhaps by switching jade=>jade-legacy we can safely remove the shrinkwrap as the security patches wouldn't be problematic anymore. /cc @nickbalestra

PR #496, opened 2017-05-31
Node 8

matteofigus:
🎉 🎈 😃


matteofigus:
Looks like 🔴 because of npm install. 
Let's keep it opened and will try to reproduce locally

diffalot:
I have been able to reproduce this.

https://github.com/opentable/oc/blob/master/npm-shrinkwrap.json#L5066-L5070

it turns out that npm 3.10.8 has a dependency on asap that resolves to an uber internal registry, and installing on node 6 with npm 3.10.10 works fine, but installing with node 8 and npm 5 does not work.

I'm not sure what the path past this is, it may be resolved by #404 or perhaps a upgrade to a newer npm version is safe to do?

running `npm install --save npm` seems to get around the error

matteofigus:
Thanks for digging into this @diffalot, perhaps a quick upgrade to latest npm could do the job 👍

mattiaerre:
shall we do something like:

```yaml
before_install:
  - "npm install -g npm@5"
```

in the .travis.yml file?

matteofigus:
@mattiaerre I think the issue is not to the npm used for the install, but the npm used as dependency in the registry. I will update that asap and it should fix. Leave it to me ;)

matteofigus:
Closing this. Will reopen after I get it green locally and after merging #497

Issue #298, opened 2016-09-30
Migrate Jade to Pug


matteofigus:
This will fix #297 and in addition will allow some cleanup in terms of the npm-shrinkwrap overrides.

nickbalestra:
Note: once we have the plugin template plugin system in place, create a pug template plugin to support it.

matteofigus:
I am closing this, because the new plan is to instead remove the jade template from the base templates with a new breaking change in the future. In that scenario, people will be able to just use the new template for pug.

PR #497, opened 2017-06-01
Fix vulnerabilities and remove npm shrinkwrap

### LINKED ISSUE #404: get rid of npm-shrinkwrap

matteofigus:
Fixes #404 and addresses a little bit of #298 (jade to pug migration is internal, but still not supported by templates).

After this upgrades, Snyk should go green so the shinkwrap should be not required anymore.


package.json
```diff
@@ -68,7 +68,8 @@
     "form-data": "0.1.4",
     "fs-extra": "0.30.0",
     "infinite-loop-loader": "1.0.1",
-    "jade": "1.11.0",
+    "jstransformer-clean-css": "^2.0.0",
+    "jstransformer-uglify-js": "^1.2.0",
```
nickbalestra: As I don't know much about those libraries, where/how do those helpers being registered/used?  (I guess you using the clean-css [here](https://github.com/opentable/oc/pull/497/files#diff-52280d37108c931558964db275372680R18), but couldn't find other references to those in the PR)

matteofigus: Jade has bultin transformers for uglify-js and clean-css. With Pug, they modularised all the tranformers and you need to install all the transformers you want to use separately. That's why I had to include them. Also, 'uglify-css' has been renamed to 'clean-css'.

PR #500, opened 2017-06-02
Show available dependencies in the UI

### LINKED ISSUE #490: Expose whitelisted dependencies via the Registry UI
With available versions and differentiation between native/non native

matteofigus:
Fixes #490 

Each dependency will be linked to either the package homepage or the node docs page (in case of core dependency).

screen shot 2017-06-02 at 23 49 22

### DIFF
--- a/src/registry/views/components-dependencies.pug
+++ b/src/registry/views/components-dependencies.pug
@@ -1,0 +1,9 @@
+#components-dependencies.box
+  each dependency in availableDependencies
+    .componentRow.row.table
+      a(href=dependency.link, target="_blank")
+        if dependency.core
+            p.release #{dependency.name} (node.js core dependency)
+        else
+            p.release #{dependency.name}@#{dependency.version}
+        
--- a/src/registry/views/components-dependencies.pug
+++ b/src/registry/views/components-dependencies.pug
@@ -6,4 +6,3 @@
             p.release #{dependency.name} (node.js core dependency)
         else
             p.release #{dependency.name}@#{dependency.version}
-        
--- a/src/registry/routes/helpers/get-available-dependencies.js
+++ b/src/registry/routes/helpers/get-available-dependencies.js
@@ -1,0 +1,17 @@
+'use strict';
+
+const coreModules = require('builtin-modules');
+const _ = require('lodash');
+
+const RequireWrapper = require('../../domain/require-wrapper');
+
+module.exports = (dependencies) => _.map(dependencies, (dependency) => {
+
+  const requirer = RequireWrapper(dependencies);
+  const core = _.includes(coreModules, dependency);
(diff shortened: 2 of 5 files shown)


src/registry/views/components-dependencies.pug
```diff
@@ -0,0 +1,9 @@
+#components-dependencies.box
+  each dependency in availableDependencies
+    .componentRow.row.table
+      a(href=dependency.link, target="_blank")
+        if dependency.core
+            p.release #{dependency.name} (node.js core dependency)
+        else
+            p.release #{dependency.name}@#{dependency.version}
+
```
mattiaerre: add `newline` please

src/registry/views/components-dependencies.pug
```diff
@@ -0,0 +1,9 @@
+#components-dependencies.box
+  each dependency in availableDependencies
+    .componentRow.row.table
+      a(href=dependency.link, target="_blank")
+        if dependency.core
```
mattiaerre: do you think we should list core dependencies?

matteofigus: Core deps are subject to be whitelisted too, so I think it's useful to show them. You can see 'url' in the screenshot example

src/registry/views/components-dependencies.pug
nickbalestra: Good catch @mattiaerre , what you think of adding prettier in the near future so that we don't have much nits anymore

src/registry/views/components-dependencies.pug
mattiaerre: ok, I see. I thought they were implicitly available for us 😊

src/registry/views/components-dependencies.pug
mattiaerre: I'd love to have that @nickbalestra 👍

mattiaerre:
Happy to approve if you answer my main question here:

https://github.com/opentable/oc/pull/500#pullrequestreview-41946918

thanks!

src/registry/views/components-dependencies.pug
matteofigus: ok done

matteofigus:
@mattiaerre I would be ok on having an api endpoint for that, but I would do a separate PR.

Perhaps we can open an issue first to discuss, as I have opinions to discuss in regards of versioning, the url structure, authentication, and the endpoint scope (one for deps? or one generic with all the details)?

mattiaerre:
@matteofigus very good then, as @corlobepy and I would like to work on a list of plugins as well and I was more inclined to expose them via an API so I was thinking that maybe we can have an API endpoint for deps as well. are you happy if we create a couple of issues and PRs on this particular subject? thanks

matteofigus:
Sure let's create an issue and discuss there ;)

Issue #464, opened 2017-04-12
get rid of grunt


matteofigus:
Replace it with npm tasks and then make the custom tasks in the /task folder to be just node tasks

jakeclements:
I'd be happy to take a crack at removing grunt as my first contribution if no one else is working on it

nickbalestra:
Hey Jake, all good and that will be awesome. Thanks for wanting to help on this!

PR #507, opened 2017-06-05
Pretty printed code

nickbalestra:
Initial PR to check out prettier.

The following PR use prettier-eslint to pipe code through prettier->eslint--fix to format code. Meaning that eslint will be run last allowing for a final override of the prettier output.

I see we have quite a big eslintrc file, I guess it could be a bit more minimal perhaps? What do you think @elboletaire ?

Following steps could be:
- use `lint-staged` together with `husky` for a totally invisible/painless workflow. I've been using it with prettier as suggested and its pretty awesome.

@michelebertoli would love to hear your thoughts in hooking prettier with eslint in here

cc @mattiaerre @matteofigus

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -2,16 +2,15 @@
 
 const _ = require('lodash');
 
-module.exports = function(grunt){
-
+module.exports = function(grunt) {
   const taskObject = { pkg: grunt.file.readJSON('package.json') };
 
-  grunt.file.expand('tasks/*.js', '!tasks/_*.js').forEach((file) => {
+  grunt.file.expand('tasks/*.js', '!tasks/_*.js').forEach(file => {
     let name = file.split('/');
     name = name[name.length - 1].replace('.js', '');
     const task = require('./' + file);
 
-    if(_.isFunction(task)) {
+    if (_.isFunction(task)) {
       task(grunt);
     } else {
       taskObject[name] = task;
@@ -22,10 +21,14 @@ module.exports = function(grunt){
   require('load-grunt-tasks')(grunt);
 
   grunt.registerTask('default', ['test-local', 'build']);
-  grunt.registerTask('sauce', ['karma:sauce-linux', 'karma:sauce-osx', 'karma:sauce-windows']);
-  grunt.registerTask('test-local', ['eslint', 'mochaTest:all', 'karma:local']);
-  grunt.registerTask('test-local-silent', ['eslint', 'mochaTest:silent', 'karma:local']);
-  grunt.registerTask('test', ['eslint', 'mochaTest:all']);
+  grunt.registerTask('sauce', [
+    'karma:sauce-linux',
+    'karma:sauce-osx',
+    'karma:sauce-windows'
+  ]);
+  grunt.registerTask('test-local', ['mochaTest:all', 'karma:local']);
+  grunt.registerTask('test-local-silent', ['mochaTest:silent', 'karma:local']);
+  grunt.registerTask('test', ['mochaTest:all']);
   grunt.registerTask('git-stage', [
     'gitadd:versionFiles',
     'gitcommit:version',
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "chai": "3.5.0",
     "cheerio": "0.22.0",
     "grunt": "^1.0.1",
-    "grunt-eslint": "^19.0.0",
     "grunt-git": "^1.0.0",
     "grunt-karma": "^2.0.0",
     "grunt-mocha-test": "^0.12.7",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,14 @@
     "oc": "./src/oc-cli.js"
   },
   "scripts": {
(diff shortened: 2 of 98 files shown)


nickbalestra:
@matteofigus yes, just wanted to make sure we all liked prettier output first :) 

TODO:
- [ ] `husky` + `lint-staged`

nickbalestra:
@mattiaerre @matteofigus  one thing that we should also remove is running eslint before testing now, as we will automatically run it with --fix together with prettier. Thoughts?

matteofigus:
eslint in CI checks with a clean npm install, which is something I can't guarantee the user has when doing the commit. I think we should leave it here, as it's cheap and free to use.

nickbalestra:
Ideal workflow:

1. write code
2. make tests greens (a `;` shouldn't be the blocker for the test as this get into your way) <- we could remove the eslint check here
3. commit and push (the 'magic' happens on pre-commit, where all the .js file staged get piped through prettier --write -> eslint--fix and are re `git added` before commit)

The user shouldn't worry to much about formatting the code according to your opinions, and you want to be sure that everything is commited as you wanted. As you are running eslint --fix, testing against it shouldn't be necessary anymore. That's why I think we could remove it from the grunt test task.

@kentcdodds am I missing something or does this make sense ? I'm trying to understand how we could add prettier in our workflow here...

kentcdodds:
Sounds right to me. Important to note that if you want to do a partial commit of a file you'll want to run that with `--no-verify`, otherwise the whole file will get added via this git hook. But that's often not a problem in my work personally so I don't really care much :)

nickbalestra:
Indeed! thank you so much for the feedback and for `prettier-eslint`

elboletaire:
Hey sorry for the delay, I've been out these days. Answering what @nickbalestra asked when opening the PR, of course it can be reduced. Remember that I used the eslint generator for that first step and, as I said then, many of the rules should be verified. Specially now that you're adding prettier, as many of that rules where related to coding style.

Moreover, [eslint rules can extend other rules][extend]. So, another way to significantly reduce the eslint file size could be search for a configuration that fits our needs (or the majority of them) and extend from it. 

[extend]: http://eslint.org/docs/user-guide/configuring#extending-configuration-files

nickbalestra:
Thank you for the input @elboletaire! Feel free to suggest any thing you might consider we want to clean from them as w moving forward with this.

nickbalestra:
Once this land in master, would be nice to add OpenTable to this -> https://github.com/prettier/prettier/issues/1351

Gruntfile.js
```diff
@@ -22,10 +21,14 @@ module.exports = function(grunt){
...
 
   grunt.registerTask('default', ['test-local', 'build']);
-  grunt.registerTask('sauce', ['karma:sauce-linux', 'karma:sauce-osx', 'karma:sauce-windows']);
-  grunt.registerTask('test-local', ['eslint', 'mochaTest:all', 'karma:local']);
-  grunt.registerTask('test-local-silent', ['eslint', 'mochaTest:silent', 'karma:local']);
-  grunt.registerTask('test', ['eslint', 'mochaTest:all']);
+  grunt.registerTask('sauce', [
+    'karma:sauce-linux',
+    'karma:sauce-osx',
+    'karma:sauce-windows'
```
elboletaire: I'd encourage to add the dangling comma, as it's a good practice when using source control (adding it won't show a modified line every time we add a line to an array/object).

But this is something should be discussed, I guess (that's why I'm not setting this as a review comment).

nickbalestra: That was already like that, so didn't want to modify any of our rules with this PR. 
But totally, we could definitely discuss all of those one-by-one so please feel free to lead the effort with issues/pr in this direction so that we can discuss each of those improvements. Thanks for that!

package.json
```diff
@@ -35,16 +42,19 @@
...
     "karma": "^1.3.0",
     "karma-jasmine": "^1.0.2",
     "karma-phantomjs-launcher": "^1.0.2",
     "karma-sauce-launcher": "^1.0.0",
     "karma-sinon": "^1.0.5",
+    "lint-staged": "^3.6.0",
     "load-grunt-tasks": "^3.5.2",
     "mocha": "^3.0.2",
     "phantomjs-prebuilt": "^2.1.12",
+    "prettier-eslint-cli": "^4.0.4",
```
matteofigus: @nickbalestra now that we have prettier-eslint-cli and removed eslint directly from the grunt task, shouldn't we remove grunt-eslint as dev dependency, including its own /tasks file?

nickbalestra: Yes!

matteofigus: Unless we want to leave it as utility for development

nickbalestra: I think make sense to remove it (also help a little toward moving to a [gruntless world](https://github.com/opentable/oc/issues/464)).

You could now run `prettier-eslint` instead if you want for developement

nickbalestra:
@matteofigus @elboletaire should be good to go now.

PR #508, opened 2017-06-05
[GPT-646] Split module API

nickbalestra:
NEW TEMPLATE MODULES API/STRUCTURE

**whataver-templateName-compiler**
  // API:
  - compile(options, callback)
  // options.watch=true|false (in the future)
  // options.server = { serverOptions }
  // options.view = { viewOptions }
  // options.statics = { staticsOptions }

// Package.json
 - dependencies: whataver-templateNam

**whataver-templateName**
  // API:
  - getInfo(),
  - getCompiledTemplate(),
  - render()

***

TODO:
- [x] `require-template()` to handle requiring the specific module type via option param
- [x] refactor [oc-template-jade](https://github.com/opencomponents/oc-template-jade/pull/10)
- [x] refactor [oc-template-handlebars](https://github.com/opencomponents/oc-template-handlebars/pull/20)
- [x] publish majors versions of refactored template and update deps in OC
- [x] update test


nickbalestra:
oc-template-jade -> https://github.com/opencomponents/oc-template-jade/pull/10

src/utils/require-template.js
```diff
@@ -6,24 +6,28 @@ const path = require('path');
 const templateNotFound = 'Error requiring oc-template: "{0}" not found';
 const templateNotValid = 'Error requiring oc-template: "{0}" is not a valid oc-template';
 
-function isValidTemplate(template){
+function isValidTemplate(template, options){
   if (typeof template !== 'object') {
     return false;
   }
 
+  if (options.compiler === true) {
```
matteofigus: should this be a if(options.compiler) ? I just wander how it could be === true and also be a function :D

nickbalestra: Is not a function but a boolean that can be passed as an option at call time. need to add it [here](https://github.com/opentable/oc/blob/2485f6a623d884af311a8e4e60f0383316dad74d/src/cli/domain/package-template.js#L25)

Does make sense?

nickbalestra:
@matteofigus all green

Issue #511, opened 2017-06-06
Not able to add registries to current project with latest oc (0.37.7)


NapoleanReigns:
I am trying to add oc registries to current project after updating to latest oc (0.37.7).

------
$ oc registry add http://localhost:3000
Usage: oc registry add <registryUrl>
Options:
-h, --help Show help [boolean]
--version Show version number [boolean]
Examples:
oc registry add http://my-registry.in.my.domain/
No such command 'add'
-------

tried with node v8.0 and old releases as well.

am i missing anything here ??

matteofigus:
Hi @NapoleanReigns which Operating System are you using?
The command seems correct

mattiaerre:
having a look at this @matteofigus 

// cc @NapoleanReigns

mattiaerre:
I think that the problem is related to this `validate` function:

https://github.com/opentable/oc/blob/master/src/cli/index.js#L19-L26

it seems to me that it is not checking against the available commands when `level` is 1; I would propose something like this as a fix:

```javascript
function validate(argv, level) {
  let keys = _.keys(commands.commands);
  if (level === 1) {
    keys = _.keys(commands.commands[argv._[0]].commands);
  }
  if (argv._.length > level &&
    !_.includes(keys, argv._[level])) {
    throw new Error(format(strings.messages.cli.NO_SUCH_COMMAND, argv._[level]));
  }
  return true;
}
```

I will create a PR and add some tests as well,
thanks

matteofigus:
So, you can actually reproduce? I can't

mattiaerre:
@matteofigus yes I can; w/ `v0.37.7` a PR is on its way 😉

matteofigus:
@NapoleanReigns thanks for this and @mattiaerre thanks for the fix. I managed to reproduce, and this should fix. Just published it as `0.37.8`

NapoleanReigns:
@matteofigus @mattiaerre Thanks much for quick turnaround. You guys are awesome.

PR #513, opened 2017-06-06
issue w/ registry add command

mattiaerre:
# Description

it looks like `registry` is the only command that has sub-commands; the `validate` function doesn't take into account this 2nd level when checking against available commands; I extracted the function in a separate file and added some tests.


mattiaerre:
this will fix #511

PR #515, opened 2017-06-08
Ignore the .git folder for the dev watcher

Kimtaro:


src/cli/domain/watch.js
```diff
@@ -14,8 +14,9 @@ module.exports = function(dirs, baseDir, changed) {
       (fileName, currentStat, previousStat) => {
         if (!!currentStat || !!previousStat) {
           if (
-            /node_modules|package.tar.gz|_package|\.sw[op]/gi.test(fileName) ===
-            false
+            /node_modules|package\.tar\.gz|_package|\.sw[op]|\.git/gi.test(
+              fileName
+            ) === false
```
mattiaerre: just a small comment: interesting indentation 😅

matteofigus: is it prettier? :D @nickbalestra can we hack around prettier rules to make that...prettier?

mattiaerre: @matteofigus this is not `prettier` this is `figus-ier` 😄

Kimtaro: I prefer this prettified version over the original since it puts the `===` on the same line as the `false`.

matteofigus:
@Kimtaro thanks for your quick PR 👍 I just published oc@0.37.9 that contains this change

PR #516, opened 2017-06-11
Improve AWS upload mechanism and allow to override registry timeout settings

### LINKED ISSUE #514: Not able to publish components of size > 500MB
I am not able to publish a component (which is of size 570 MB) to AWS s3 bucket from local computer (Mac OS X Sierra).

I often see the below error when it fails,

**An error happened when publishing the component: Error: socket hang up**

I tried to publish the component in below manner,

---------------
node index.js bucket=<AWS-s3-bucket-location> region=<region> key=<key> secret=<secret>
oc registry add http://localhost:3000
oc publish <comonent-name>
---------------
 
Output:
Packaging -> goes fine
Compressing -> goes fine
Publishing -> component url
**An error happened when publishing the component: Error: socket hang up**

any workaround for this ??

matteofigus:
Fixes #514 

This Pull Request aims to address some necessary optimisations to the overall publish strategy, specifically with dealing with big components (including many and/or big files) and possibly slow connections.

In this perspective, here
1) We change the strategy for uploading to S3 from `putObject` to `upload` method. This allows uploads into multiple chunks and is necessary for files that reach big dimensions
2) We make customisable the default timeout of the API, at the moment set to 2m, so that when a publish takes a long time (because of the many requests to S3) the request is not terminated and handled correctly.

See inline comments for details about the implementation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
   "dependencies": {
     "accept-language-parser": "1.4.0",
     "async": "2.4.1",
-    "aws-sdk": "2.62.0",
+    "aws-sdk": "2.67.0",
     "babel-core": "6.21.0",
     "babel-loader": "7.0.0",
     "babel-preset-env": "1.1.8",
--- a/src/registry/domain/options-sanitiser.js
+++ b/src/registry/domain/options-sanitiser.js
@@ -9,17 +9,15 @@ module.exports = function(input) {
   const options = _.clone(input);
 
   if (!options.publishAuth) {
-    options.beforePublish = function(req, res, next) {
-      next();
-    };
+    options.beforePublish = (req, res, next) => next();
   } else {
     options.beforePublish = auth.middleware(options.publishAuth);
   }
 
   if (!options.publishValidation) {
-    options.publishValidation = function() {
-      return { isValid: true };
-    };
+    options.publishValidation = () => ({
+      isValid: true
+    });
   }
 
   if (!options.prefix) {
@@ -50,6 +48,7 @@ module.exports = function(input) {
     .map(s => s.toLowerCase());
 
   options.port = process.env.PORT || options.port;
+  options.timeout = options.timeout || 1000 * 60 * 2;
 
   return options;
 };
--- a/src/registry/domain/s3.js
+++ b/src/registry/domain/s3.js
@@ -148,13 +148,7 @@ module.exports = function(conf) {
 
           putFile(file, url, relativeFile === '/server.js', cb);
         },
-        errors => {
-          if (errors) {
-            return callback(_.compact(errors));
-          }
-
-          callback(null, 'ok');
-        }
+        callback
       );
     });
   };
@@ -177,17 +171,17 @@ module.exports = function(conf) {
     if (fileInfo.gzip) {
       obj.ContentEncoding = 'gzip';
     }
-
-    getClient().putObject(obj, callback);
+    const upload = getClient().upload(obj);
+    upload.send(callback);
   };
 
   const putFile = (filePath, fileName, isPrivate, callback) => {
-    fs.readFile(filePath, (err, fileContent) => {
-      if (err) {
-        return callback(err);
-      }
-      putFileContent(fileContent, fileName, isPrivate, callback);
-    });
+    try {
+      const stream = fs.createReadStream(filePath);
+      putFileContent(stream, fileName, isPrivate, callback);
+    } catch (e) {
+      callback(e);
+    }
   };
 
   return {
--- a/test/unit/registry-domain-options-sanitiser.js
+++ b/test/unit/registry-domain-options-sanitiser.js
@@ -1,22 +1,30 @@
 'use strict';
 
 const expect = require('chai').expect;
+const _ = require('lodash');
 
 describe('registry : domain : options-sanitiser', () => {
-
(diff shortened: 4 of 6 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
   "dependencies": {
     "accept-language-parser": "1.4.0",
     "async": "2.4.1",
-    "aws-sdk": "2.62.0",
+    "aws-sdk": "2.67.0",
```
matteofigus: A bug was preventing the upload() S3 function to properly upload all the chunks

src/registry/domain/options-sanitiser.js
```diff
@@ -50,6 +48,7 @@ module.exports = function(input) {
     .map(s => s.toLowerCase());
 
   options.port = process.env.PORT || options.port;
+  options.timeout = options.timeout || 1000 * 60 * 2;
```
matteofigus: This is the timeout for the http server, after which a request gets terminated with a "ECONNRESET".
When trying to upload a big file, I changed this 10m and worked fine. I opted for a 2m default (which is the node default) but left customisable by the user.

src/registry/domain/s3.js
```diff
@@ -148,13 +148,7 @@ module.exports = function(conf) {
...
           putFile(file, url, relativeFile === '/server.js', cb);
         },
-        errors => {
-          if (errors) {
-            return callback(_.compact(errors));
-          }
-
-          callback(null, 'ok');
-        }
+        callback
```
matteofigus: async.each actually fires the error as soon as it gets the first. This means that the "compact" is not necessary as the error is just 1. When testing locally, I was seeing a error with a `[Object object]` result here prior of this fix in case of S3 issues.

src/registry/domain/s3.js
```diff
@@ -177,17 +171,17 @@ module.exports = function(conf) {
     if (fileInfo.gzip) {
       obj.ContentEncoding = 'gzip';
     }
-
-    getClient().putObject(obj, callback);
+    const upload = getClient().upload(obj);
+    upload.send(callback);
```
matteofigus: For large files, the recommended approach is to use upload over putObject. Upload automatically processes the req into multiple chunks and its `Body` property accepts both string (making this change retrocompatible) and streams (see change in the `putFile` method).

src/registry/domain/s3.js
```diff
@@ -177,17 +171,17 @@ module.exports = function(conf) {
...
   const putFile = (filePath, fileName, isPrivate, callback) => {
-    fs.readFile(filePath, (err, fileContent) => {
-      if (err) {
-        return callback(err);
-      }
-      putFileContent(fileContent, fileName, isPrivate, callback);
-    });
+    try {
+      const stream = fs.createReadStream(filePath);
+      putFileContent(stream, fileName, isPrivate, callback);
```
matteofigus: Here we just create a stream reader instead of reading the whole content. For big files, this is actually necessary for not blocking the event loop with large GCs operations.

test/unit/registry-domain-options-sanitiser.js
```diff
@@ -1,22 +1,30 @@
...
+      hotReloading: false,
+      verbosity: 0,
+      customHeadersToSkipOnWeakVersion: [],
+      timeout: 120000
+    };
+
+    _.each(defaults, (value, property) => {
+      it(`should set ${property} to ${JSON.stringify(value)}`, () => {
+        expect(sanitise(options)[property]).to.eql(value);
+      });
```
matteofigus: I wanted to test `timeout` returning the default, but I just ended up testing each property of the snapshot. When migrating to `jest`, this will be so much prettier ❤️

test/unit/registry-domain-options-sanitiser.js
```diff
@@ -25,39 +33,39 @@ describe('registry : domain : options-sanitiser', () => {
...
+        customHeadersToSkipOnWeakVersion: ['header1', 'HeAdEr-TwO', 'HEADER3']
+      };
 
       it('should convert the array elements to lower case', () => {
-        expect(sanitise(options).customHeadersToSkipOnWeakVersion).to.be.eql(['header1', 'header-two', 'header3']);
+        expect(sanitise(options).customHeadersToSkipOnWeakVersion).to.be.eql([
+          'header1',
+          'header-two',
+          'header3'
+        ]);
```
matteofigus: This seems to be a prettier initiative.

src/registry/domain/s3.js
```diff
@@ -177,17 +171,17 @@ module.exports = function(conf) {
...
 
   const putFile = (filePath, fileName, isPrivate, callback) => {
-    fs.readFile(filePath, (err, fileContent) => {
-      if (err) {
-        return callback(err);
-      }
-      putFileContent(fileContent, fileName, isPrivate, callback);
-    });
+    try {
+      const stream = fs.createReadStream(filePath);
```
matteofigus: This is now inside a try/catch because a exception will be fired in case of file not found.

src/registry/domain/options-sanitiser.js
nickbalestra: Why not getting `server.timeout` instead of `1000 * 60 * 2` ?

nickbalestra: Why not relying on `server.timeout` instead of `1000 * 60 * 2` ?

PR #521, opened 2017-06-13
[GPT-728] Full compilation via Template Unified Compiler

nickbalestra:


src/cli/domain/package-components.js
```diff
@@ -2,131 +2,57 @@
...
-            return cb(null, component);
-          }
-
-          packageServerScript(
-            {
-              componentPath: componentPath,
-              dependencies: component.dependencies,
-              ocOptions: component.oc,
-              publishPath: publishPath,
-              verbose: options.verbose
```
matteofigus: we have a verbose here...

src/cli/domain/package-components.js
```diff
@@ -2,131 +2,57 @@
...
-            err => cb(err, component)
-          );
+    const type = componentPackage.oc.files.template.type;
+    const compileOptions = {
+      publishPath,
+      componentPath,
+      componentPackage,
+      ocPackage,
+      minify
+      // TODO: verbose,
```
matteofigus: ...so, should we start passing it through, even if the current compilers will ignore it?

nickbalestra: I would say yes

nickbalestra:
Passed `verbose` to it. Thanks for the suggestion.

Issue #281, opened 2016-08-18
oc.require() triggers callback when dependency is still not loaded


victorborshak:
When debugging some components on a really slow connection (throttled down to GRPS with disabled cache in Chrome) I noticed that oc.require() triggers the callback in around 2 seconds regardless of whether or not the dependency has been loaded.
I believe this happens because of hardcoded timeout value of 7e3 in head.load.js
https://github.com/opentable/oc/blob/master/src/components/oc-client/src/head.load.js#L501

I believe this value is too small and should be increased to at lease 30 seconds (or a minute) or so.

matteofigus:
Hi @victorborshak thanks for your feedback. We use head.js for that and I am not happy about the fact that it seems not really up-to-date and actively maintained.

In that regards, I would prefer, instead of hacking on the head.js, to evaluate a migration to a better supported dependency loader system.

I personally love [SystemJS](https://github.com/systemjs/systemjs) but I would like other contributors' opinion about this, perhaps @mattiaerre @nickbalestra @maxgurewitz @antwhite @chriscartlidge

I list here some of the currently supported functionalities, that we may want to re-discuss:
- Possibility to both load JS or CSS with a callback
- Support for IE8
- Small so that it can be part of the `client.js` without impacting significantly to its size
- Licensed with MIT or compatible OSS license

matteofigus:
Hi @victorborshak before working on this I would like to have a working failing example so that we can use it - in case - as a foundation for the new system we are gonna put in place to solve the issue.

Would you be able to share some code with a reproducing failing scenario?

navamgupta:
@matteofigus how about using lightningjs 

https://github.com/olark/lightningjs/blob/master/README.md

We run into this problem way too often with conflict in libraries or variables in global scope. This could help with that.

matteofigus:
@navamgupta lightningjs seems neat, but unfortunately I see no activity on github for the last 5 years - which is practically the same problem I am trying to solve now by replacing head.js.

I think I would be happy to test a couple of tools, but if we could get an example in github, even a simple gist with a conflict example with reproduction steps, trying out different tools would facilitate the problem's resolution as we could use the bugs as foundation to write some tests on the new integration.

navamgupta:
[...]

PR #522, opened 2017-06-20
Issue w/ head.load

mattiaerre:
# Description

the following gif shows that when replacing `head.load` w/ `l` it is possible to wait until the asset has been loaded before firing a callback function; this PR includes also the 2 html files used in order to record the video.

the first part shows the behavior w/ `head.load` simulating a Regular 2G network w/ Chrome dev tool; the second part shows the behavior w/ `l` simulating the same network speed.

The test is fetching a `.css` and a `.js` file.

ljs-2

the [source code](https://github.com/malko/l.js) of the `l` library 

## Changes

- add the `l` library (and update .eslintignore)

- update `build.js` so that it is including `l` inside `oc-client.min.js` when `grunt build`

- use `l` instead of `head.load` inside karma-common (grunt karma:local still works)


mattiaerre:
this PR tries to fix #281 

// cc @matteofigus

tasks/build.js
```diff
@@ -7,46 +7,81 @@ const uglifyJs = require('uglify-js');
...
-        componentPath: path.join(__dirname, clientComponentDir),
-        verbose: false
-      };
-
-    local.package(packageOptions, (err) => {
-      grunt.log[err ? 'error' : 'ok'](err ? err : 'Client has been built and packaged');
-      done();
-    });
-  });
+module.exports = function(grunt) {
```
mattiaerre: this has been reformatted because of `prettier`

src/components/oc-client/src/l.js
```diff
@@ -0,0 +1,152 @@
...
+      for (i = 0, l = scripts[length]; i < l; i++) {
+        (url = scripts[i].getAttribute('src')) && (loaded[url.replace(/#.*$/, '')] = true);
+      }
+      links = D[getElementsByTagName]('link');
+      for (i = 0, l = links[length]; i < l; i++) {
+        (links[i].rel === 'stylesheet' || links[i].type === 'text/css') && (loaded[links[i].getAttribute('href').replace(/#.*$/, '')] = true);
+      }
+    }
+    //export ljs (as `head`)
+    window.head = loader;
```
mattiaerre: this is the only `workaround` applied 😅

matteofigus: Well this is temporary right? If we would want to go for it, we can change the library to make it use ljs instead and it should work, right?

matteofigus:
C'est formidable!

I think that this library does the job well, and despite not being super active, it makes all the tests pass, so I would say let's go for it.

I see that you did a good job on setting up a manual test too. The only think we may want to test (which I think is not covered by the karma tests) is css loading with oc.require. This library should support it according to the docs, but I just to ensure that we can keep all retro-compatible ;)

src/components/oc-client/src/l.js
mattiaerre: do you want me to use `ljs` instead of `head` as the name of the loader? sure I'll try to do that. I will also test CSS loading.

mattiaerre:
@matteofigus I updated this PR that now uses a global `ljs` name instead of `head`; I've also fixed the tests and tried to load a `.css` file (and re-recorded a short video w/ that). the only caveat w/ css is that the callback is raised synchronously before the load completion. But this happens for both `head.load` and `ljs`. It is how CSS loading works (I guess). If you are happy I can get rid of the examples and the `head.load` lib from this PR so that you can merge and build; unless you want me to include the build here.

src/components/oc-client/src/l.js
```diff
@@ -0,0 +1,152 @@
...
+      for (i = 0, l = scripts[length]; i < l; i++) {
+        (url = scripts[i].getAttribute('src')) && (loaded[url.replace(/#.*$/, '')] = true);
+      }
+      links = D[getElementsByTagName]('link');
+      for (i = 0, l = links[length]; i < l; i++) {
+        (links[i].rel === 'stylesheet' || links[i].type === 'text/css') && (loaded[links[i].getAttribute('href').replace(/#.*$/, '')] = true);
+      }
+    }
+    //export ljs
+    window.ljs = loader;
```
mattiaerre: as per the lib now the global name is `ljs`

matteofigus:
Sounds good. Let's remember to update this file too https://github.com/opentable/oc/blob/master/src/components/oc-client/LICENSES

src/components/oc-client/LICENSES
```diff
@@ -27,7 +27,7 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 OC wouldn't exist without some fantastic Open-Source libraries:
 
 * jQuery
-* head.js
+* l.js
```
mattiaerre: @matteofigus let me know if this is good for you

mattiaerre:
all 💚  over to you @matteofigus

matteofigus:
Fantastic. I decided to go live with a new minor (0.38.0) because despite this should be a non breaking change, there was a non explicit contract on exposing `head` so it is better to toggle a minor for clarity. For most of the user, this should be an easy upgrade.

mattiaerre:
I wanted to wait a couple of weeks (just to make sure that there was no production issue) but now I'd like to take some time and thank @malko for his `l.js` library that helped us smoothly to transition from `head.js`. Very good job and thanks again; OSS rulez 😎 

malko:
@mattiaerre thank for your mention and use of ljs. Glad it can help you. As said earlier in the comment the repository is not really active but i want to assure you that this is not a dead project, I still accept PR and take a look to it on a regular basis. I just don't have more need for it at this time but as said PR are welcome as long as it don't forget the purpose of ljs, stay a micro library.
Hope you'll be happy with it.

PR #528, opened 2017-07-04
[OC-50] Log body fix

matteofigus:
When looking at the request event, and the request body's object, that didn't seemed to be the original request body.

After some digging, I found out what the cause was. This is the fix.

I also did some testing and changed some dependencies in order to allow my local registry to run without problems.

### DIFF
--- a/src/registry/routes/components.js
+++ b/src/registry/routes/components.js
@@ -50,18 +50,18 @@ module.exports = function(conf, repository) {
       components,
       (component, callback) => {
         getComponent(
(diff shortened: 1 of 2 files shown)


src/registry/routes/components.js
```diff
@@ -50,18 +50,18 @@ module.exports = function(conf, repository) {
       components,
       (component, callback) => {
         getComponent(
-          _.extend(component, {
```
nickbalestra: +1

PR #532, opened 2017-07-07
fix-global-ljs

mattiaerre:
# Description

- fix references to `head` replaced w/ `ljs`


nickbalestra:
Merging this here and then bringing it over to https://github.com/opencomponents/oc-client-browser

PR #538, opened 2017-07-11
[OC-69] Remove node client

nickbalestra:
Blocking PRs:
- [x] https://github.com/opencomponents/oc-client-node/pulls need to be published before this PR could work


src/index.js
```diff
@@ -1,4 +1,3 @@
 'use strict';
 
-module.exports.Client = require('../client');
```
matteofigus: I think I would keep support for this, just by exposing `require('oc-client')`

PR #542, opened 2017-07-13
Drop node4 tests

### LINKED ISSUE #437: Remove support for Node 4 and 5 for oc core
So we can start supporting only 6, 7 (and very soon 8).
(node oc-client would still support 4 and 5).

This is for letting anyone know we are planning to do this soon. Any objections?

nickbalestra:
Context: The node client has been moved out to [his own repo](https://github.com/opencomponents/oc-client-node) and it still support node4. This allows us to now drop support for node4 on the cli/registry where we support `>=6`

Fixes #437

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: node_js
 
 matrix:
   include:
-    - node_js: "4"
-    - node_js: "5"
     - node_js: "6"
     - node_js: "7"
     - node_js: "8"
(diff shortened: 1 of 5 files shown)


nickbalestra:
If run on node below v6 the following alert will be displayed in the terminal (example running 4.8.3):

```bash
ALERT: You're currently running OC on an unsupported Node version (v4.8.3).
Please upgrade Node to >= 6.0.0.
```

Issue #540, opened 2017-07-13
mime type for html files


NapoleanReigns:
I am trying to push a component with some html files in it. 
The mime type gets pushed to the library as `application/octect-stream`.

Can we add a mime type of html files as well.

I see the mime types defined in the below file,
https://github.com/opentable/oc/blob/master/src/utils/get-mime-type.js

matteofigus:
sure @NapoleanReigns - would you like to make a pull request, or want us to do it?

NapoleanReigns:
@matteofigus i can float a pull request.

PR #545, opened 2017-07-14
Added a mime type for html files.

NapoleanReigns:
I have added a mime type for html files as `text/html`.

If this is not defined at oc level and you are trying to serve the files over a front facing ngnix server (on top of oc) in a cloud environment - it would ultimately pick up the default mime type defined in ngnix which is `application/octet-stream`.

Let me know if you see any concerns here with the commit.


matteofigus:
Fixes #540

PR #583, opened 2017-07-30
Saner registry default for config.discovery

nickbalestra:
the discovery option was not documented. I've updated the wiki with it.
I suggest we set it on true by default so that is friendlier for everybody starting with OC and who follows the getting started for setting up a registry.


matteofigus:
LGTM. Will publish as breaking.

PR #595, opened 2017-08-05
Add ability to disable watching on local dev registry

badsyntax:
For our use case, we run visuals tests against the local dev registry. This process saves screenshots to disk which then kicks off an OC package task and sometimes this errors out whilst the visual tests are running. We'd love to be able to disable watching for this scenario.

Would need to update: https://github.com/opentable/oc/wiki/Cli

### DIFF
--- a/src/cli/commands.js
+++ b/src/cli/commands.js
@@ -27,6 +27,11 @@ module.exports = {
           boolean: true,
           description: 'Verbosity',
           default: false
+        },
+        watch: {
+          boolean: true,
+          description: 'Watch for file changes',
+          default: true
         }
       },
       usage: 'Usage: $0 dev <dirName> [port] [baseUrl] [options]'
--- a/src/cli/facade/dev.js
+++ b/src/cli/facade/dev.js
@@ -25,7 +25,8 @@ module.exports = function(dependencies) {
       fallbackRegistryUrl = opts.fallbackRegistryUrl,
       hotReloading = _.isUndefined(opts.hotReloading)
         ? true
-        : opts.hotReloading;
+        : opts.hotReloading,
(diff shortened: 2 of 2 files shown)


badsyntax:
Actually it looks like `--hotReloading=false` is good enough, so i'm gonna close this one.

nickbalestra:
@badsyntax I see you reopened this PR, any more context to share?

badsyntax:
Hi @nickbalestra 

`--hotReloading=false` gets us most of the way but i still feel it's a workaround.

For our use case we simply don't want any watching at all. Even with `--hotReloading=false`, the watcher still monitors every file (including `node_modules`) and this can be a pretty intensive task. We want to run a dev registry, with production built files, and then perform some tasks on that registry (like do visual regression tests), we have no need for watching, and it eats up CPU and disk unnecessarily. 

In some cases it results in obscure errors, like `ESOCKETTIMEDOUT` when doing network requests. We think this is related to open file limits (see "Open File Limits" here: http://kb.imakewebsites.ca/2014/01/03/nodejs-setup-tips-for-devops/), but we're not sure. Multiple people in our team have had this error, and sure, this is more of a generic problem rather than a problem with OC, but it would be still be useful to be able to disable this behaviour in OC.

ryaanwells:
Seconding @badsyntax on this one 👍 

Just to add one more case this would be useful in - consider when the OC component loads some ES6 JS, scss or React code that needs transpiling - you need to rebuild entirely to see the changes (hot reloading the dev registry won't help). I want to control the lifecycle of the registry reload as part of my component rebuild lifecycle and would rather it didn't do anything in isolation.

nickbalestra:
@badsyntax published with `v0.40.6`, and [added documentation](https://github.com/opentable/oc/wiki/Cli#options)

Issue #473, opened 2017-04-24
OC Registry needs ❤️ (and tests)


mattiaerre:
# Description

We noticed that the OC Registry code is very low on tests, this issue is in order to plan some activities for testing and improving that codebase.

matteofigus:
I am not a big fan of test coverage as a goal, but I think it would be useful, every time we add features or we do housekeeping PRs, to visualise that the coverage % at least doesn't decrease over time. What's your thoughts on that? Perhaps we could add a reporter that shows the coverage as a badge, to motivate us contributors (even first time contributors) to make that number slowly increase.

mattiaerre:
@matteofigus I think we've already touched this subject and I'm 100% in line w/ you: code coverage is not a goal but it can be a very useful metric to understand where the problems may be; one of my fav quotesxt:

> Although 100% code coverage may appear like a best possible effort, even 100% code coverage is estimated to only expose about half the faults in a system.
> Low code coverage indicates inadequate testing, but high code coverage guarantees nothing.

Tomorrow I am going to have a conversation w/ @pbazydlo @nickbalestra @Kimtaro and @chriscartlidge about testing w/ Jest and I think it will be very beneficial to use it also in this context; I will try to elaborate more this concept w/ a simple POC; also remember that there is a PR still outstanding for you to review? 😊 

https://github.com/opentable/oc-testing/pull/1

nickbalestra:
+1 on Jest as it provide coverage

nickbalestra:
@mattiaerre, would u like to start attempting a PR in order to move to jest? Would love to help on it in order to make this happens!

mattiaerre:
@nickbalestra definitely. What about moving the registry 's code to the opencomponents organization first? I will create a new repo and see what happens 😉

nickbalestra:
I'm more in favor to make it into his own package within the OC repo for the moment (AKA going monorepo), later we can then decide if we want to add `lerna`/`northbrook` to handle the monorepo, keep our own scripts or simply move it to its own repo. Does this make sense to you? @matteofigus any thoughts in this direction?

matteofigus:
My personal preference is to go mono-repo, but I would like to talk about it first. I'll create an issue.

mattiaerre:
@matteofigus @nickbalestra fine by me but from a testability POV I am not sure if we can have only the Registry part w/ Jest. Let me have a look and see if there is anything that can help us w/ a partial migration 😉

[...]

PR #597, opened 2017-08-05
[OC-120] Registry webUI: preview route/view bug

nickbalestra:
Preview route didnt take into account of registry prefix for paths


matteofigus:
The baseUrl should be set to contain the prefix too. Perhaps we should improve the baseUrl docs?

nickbalestra:
Mmmm I see...
Yes totally not clear.
My personal feeling is that if you use two seperate configuration then the two things should be indipendent, no?

nickbalestra:
I have an idea 😎 , let me rework this PR

nickbalestra:
@matteofigus I've extended the `options-sanitizer`:
It now checks if a prefix exist (before sanitizing it) and if the baseUrl doesn't already contain the prefix. If both check succeed it it sanitize the baseUrl by appending the prefix to it.

This should avoid this being a braking change, and will make it possible to handle both cases:
1. prefix and baseUrl containing prefix
2. prefix and baseUrl without prefix

Let me know what you think

nickbalestra:
add
- [ ] sanitize baseUrl with trailing slash as per https://github.com/opencomponents/get-started/issues/2

nickbalestra:
@matteofigus do want to review this one more time before 🌴 🌞  🏄  🍺  🏖  ?

nickbalestra:
@mattiaerre to get into merge mood, what about starting with this? 😎

nickbalestra:
There are still a couple of issues that need to be addressed that seems the tests didn't cought:
if prefix && baseUrl contiene prefix, niente duplicate prefix...
I'll need to add more tests to cover also this scenario

nickbalestra:
@mattiaerre I've rebased and added the scenario based tests for the the baseUrl/Prefix.

I'll rather keep this PR scoped on solving this very small 'bug' that we discovered with the heroku get started registry.
I therefore suggest we don't addd jest or takle other aspects of the sanitization like `beforePublish` yet. (Also no need to add jest to have snapshots as those are now avail for any test runner).

Can you open an issue for the `beforePublish`? Ffor jest we already have #473 
If Ok i'll revert the commit that added jest.

nickbalestra:
@mattiaerre this should be now ready to go.
I'll update the OpenComponents website registry and the get-started repo once this is merged. Thanks!

.gitignore
```diff
@@ -27,4 +27,4 @@ src/components/base-component-handlebars/_package
 src/components/base-component-jade/_package
 
 test/fixtures/test.tar.gz
-test/fixtures/targz-test
\ No newline at end of file
+test/fixtures/targz-test
```
mattiaerre: me love this!

test/unit/registry-domain-options-sanitiser.js
```diff
@@ -69,4 +76,48 @@ describe('registry : domain : options-sanitiser', () => {
       });
     });
   });
+
+  describe('prefix and baseUrl sanitaziation', () => {
+    const prefixAndBaseUrlSchenario = [
```
mattiaerre: love it!

mattiaerre: typo `prefixAndBaseUrlSchenario` => `prefixAndBaseUrlScenarios`

test/unit/registry-domain-options-sanitiser.js
```diff
@@ -7,7 +7,7 @@ describe('registry : domain : options-sanitiser', () => {
   const sanitise = require('../../src/registry/domain/options-sanitiser');
 
   describe('when options is empty', () => {
```
mattiaerre: this is no longer the description for this test is it?

test/unit/registry-domain-options-sanitiser.js
```diff
@@ -69,4 +76,48 @@ describe('registry : domain : options-sanitiser', () => {
       });
     });
   });
+
+  describe('prefix and baseUrl sanitaziation', () => {
```
mattiaerre: typo `sanitaziation` => `sanitization`

PR #604, opened 2017-08-13
Updated logo

nickbalestra:


nickbalestra:
@mattiaerre added a logo with type and the logo only to be used whenever needed accordingly

mattiaerre:
@nickbalestra I was more asking about the purple only logo and the one w/ square/triangle/cylinder components

nickbalestra:
In theory empty one is for the registry. But its just a logo I wouldn't be too corporate with guideline here, use them as you like. will probably create a repo with original sketch files and CC licence on the opencomponents org.

PR #605, opened 2017-08-14
[OC-102] Return rendered version if client doesn't support

nickbalestra:
Related to (and blocked by) 

- https://github.com/opencomponents/oc-client-node/pull/8
- https://github.com/opencomponents/oc-client-node/pull/9

Registry will handle server side rendering in case of oc-client requesting the unrendered version of an unsupported template

### DIFF
--- a/src/registry/routes/helpers/get-component.js
+++ b/src/registry/routes/helpers/get-component.js
@@ -147,11 +147,8 @@ module.exports = function(conf, repository) {
 
         // Support legacy templates
         let templateType = component.oc.files.template.type;
-        if (templateType === 'jade') {
-          templateType = 'oc-template-jade';
-        }
-        if (templateType === 'handlebars') {
-          templateType = 'oc-template-handlebars';
+        if (templateType === 'jade' || templateType === 'handlebars') {
+          templateType = `oc-template-${templateType}`;
         }
 
         const filterCustomHeaders = (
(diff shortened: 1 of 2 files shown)


src/registry/routes/helpers/get-component.js
```diff
@@ -145,6 +145,15 @@ module.exports = function(conf, repository) {
           });
         }
 
+        // Support legacy templates
+        let templateType = component.oc.files.template.type;
```
mattiaerre: at this point, we are positive that `component.oc.files.template` is not `undefined` right?

test/unit/registry-routes-helpers-get-component.js
```diff
@@ -96,4 +96,72 @@ describe('registry : routes : helpers : get-component', () => {
       expect(fireStub.args[0][1].status).to.equal(500);
     });
   });
+
+  describe("when oc-client request an unrendered component and it doesn't support the correct template", () => {
```
mattiaerre: very minor but I would still use single quotes and either escape the apostrophe w/ a '\' or even consider to use '`' instead. 

also the describe should be `when oc-client requests ...`

src/registry/routes/helpers/get-component.js
nickbalestra: Yes

nickbalestra:
@mattiaerre fixed

nickbalestra:
Yes, before merging I also want to be sure this is safe for usage with clients that don't yet send template info headers, perhaps we could `isClientRequest` to also match from a specific version upward?

nickbalestra:
Yes lets wait till we publish the client first as we need to update the client dependency in OC too for this to work correctly

PR #612, opened 2017-08-25
[OC-127] Fix templates breaking change

matteofigus:
This solves a bug that prevents the clients to proper dial with the registry


nickbalestra:
@matteofigus  Those are also related and need to be addressed to align with latest changes:
- https://github.com/opencomponents/oc-client-node/pull/8
- https://github.com/opencomponents/oc-client-node/pull/9

Issue #618, opened 2017-08-31
Accessing s3 over a proxy from oc-registy


NapoleanReigns:
I am hosting my oc-registry process as a seperate docker container over an EC2 instance in AWS and the access to s3 bucket from this instance would be via a proxy.
we have been patching our local version of oc with below piece of code change for this to work for our use case.

Patch File: src/registry/domain/s3.js

`const proxy = require('proxy-agent'); 
module.exports = function (conf) {
 const httpOptions = { timeout: conf.s3.timeout || 10000 }; 
 if (conf.s3.agentProxy) { httpOptions.agent = proxy(conf.s3.agentProxy); } 
AWS.config.update({ 
accessKeyId: conf.s3.key, 
secretAccessKey: conf.s3.secret, region: conf.s3.region, 
httpOptions, });`

this needs to be fixed here.
https://github.com/opentable/oc/blob/6651c60be60e46a6803e865074f8e632e65f489f/src/registry/domain/s3.js#L20

i think httpOptions should also consider conf.s3.agentProxy.

Thank you

matteofigus:
I think I would be in favour of adding this optional property. Would you be able to do a PR for that?

NapoleanReigns:
Sure, I will float a PR for considering conf.s3.agentProxy property under configuration.

Thank you.

NapoleanReigns:
I have created a PR. 
Please Review.

Thanks

NapoleanReigns:
Thanks Matteo.

PR #619, opened 2017-08-31
Issue #618 : Accessing s3 bucket over a proxy from oc-registry

NapoleanReigns:
agentProxy property defined under the registry configuration object needs to be considered while establishing a connection to S3 Bucket from oc-registry.

This will allow the user to host the oc registry over an infrastructure behind a proxy.

Thanks


src/registry/domain/s3.js
```diff
@@ -11,13 +11,19 @@ const _ = require('lodash');
 const getFileInfo = require('../../utils/get-file-info');
 const getNextYear = require('../../utils/get-next-year');
 const strings = require('../../resources');
+const proxy = require('proxy-agent');
 
 module.exports = function(conf) {
+  const httpOptions = { timeout: conf.s3.timeout || 10000 };
+  if (conf.s3.agentProxy) {
+    httpOptions.agent = proxy(conf.s3.agentProxy);
```
matteofigus: I'm a bit confused. If the property would be called `conf.s3.proxy` I would imagine it would be consistent to assign to the `httpOptions.agent` the `proxy(url)` value, but given we called it `agentProxy` I thought it was part of the idea to pass the whole agent already?

I think this would be a bit better given the naming, and also to allow who's not interested to use the proxy to avoid installing the `proxy-agent` dependency.

What's your thoughts on that @mattiaerre @NapoleanReigns

NapoleanReigns: I completely agree with it. We shouldn't mandate the proxy-agent dependency to the user unless he is in need of it.

I will make the changes and submit soon.

NapoleanReigns:
I have incorporated the review comments and pushed the changes.

PR #617, opened 2017-08-30
[OC-140] Declarative templates initialisation on a registry level

nickbalestra:
Templates declaration is now more explicit when configuring the registry (same API as per the node client):
```
conf = {
...
templates: [require('oc-template-react')]
...
}
```

PR #623, opened 2017-09-04
[INVESTIGATION] Revert "[OC-140] Declarative templates initialisation on a registry level"

nickbalestra:
Reverts opentable/oc#617
This won't allow `cli dev` command to work with components built around templates other then the preconfigured ones (jade, handlebars)


nickbalestra:
This is probably due to how we register templates upon starting the dev registry -> https://github.com/opentable/oc/blob/master/src/cli/facade/dev.js#L148

nickbalestra:
Closed in favor of https://github.com/opentable/oc/pull/624

Issue #351, opened 2017-01-13
Revisit dev watcher-logic


nickbalestra:
As per #346  there will be the need to rework the dev watcher logic. 
Now we watch on changes on 'server.js' but with the new compiler pipeline we open scenarios where changes might happen in a local module file and not directly in the main entry server file.

nickbalestra:
Perhaps we could simply rely on the watcher provided with webpack, that should not only solve the issue but due to its caching-logic should speed-up the bundling process.
https://webpack.github.io/docs/node.js-api.html#watching

matteofigus:
@nickbalestra tagging templates here as watching mechanism once templates are done could just be part of each template's configuration and in case this issue would fix.

nickbalestra:
Totally

On 11 May 2017, at 12:02, Matteo Figus <notifications@github.com<mailto:notifications@github.com>> wrote:


@nickbalestra<https://github.com/nickbalestra> tagging templates here as watching mechanism once templates are done could just be part of each template's configuration and in case this issue would fix.

debopamsengupta:
Hi guys, noticed you have this issue open and wanted to ask if there could be an improvement to the watcher to ignore changes in anything mentioned in `.gitignore` ? 

I usually always end up with 
`Changes detected on file: /<path-to-component>/.git/FETCH_HEAD`

nickbalestra:
Hey Dev, did you find any work-around/solution to this?

On a side I started doing some explorations on the watcher logic, and I think there are few small things we could quickly change to improve the overall DX.
- Atm once a change is detected we repackage [all the components ](https://github.com/opentable/oc/blob/master/src/cli/facade/dev.js#L46)in the dev registry directory, I think we could just repackage the affected component.
- Once the packaging happen upon a change we could rely on a live reload logic to automatically refresh the preview page. This should shorten the feedback loop when dev quite a bit.  What you recon?

debopamsengupta:
Hi,
That sounds good :) I can also take a look as to how we can potentially ignore certain files/folders (my thoughts are, mainly `node_modules`)
Right now, we usually disable the watcher logic if we are working with more than 1 component :(

nickbalestra:
This should address the first point -> https://github.com/opentable/oc/pull/632. Perhaps it could help you in the scenario you describe where you have to disable the watcher if working on multiple components

danrr:
It would be nice to be able to define a white/blacklist for the watcher in, say, package.json or oc.json or as a CLI arg

[...]

PR #632, opened 2017-09-08
Improve watching for Dev registry

nickbalestra:
This PR addresses #351:

- [x] when on dev mode, only the component that was changed get re-compiled, not all the components. This should reduce feedback-loop times quite a bit when working in projects containing more components.
- [x] ignored files/dirs are now fully ignored from the tree of files to watch

Because of the combination of the two points above most of the reported issues should be address without the need for the user to explicitly define an ignore list.


nickbalestra:
Awesome, sure thing!

nickbalestra:
@matteofigus rebased

PR #635, opened 2017-09-09
liveReload

nickbalestra:
liveReload feature on the registry when in DEV mode

livereload-2

As per #351


src/registry/routes/component-preview.js
```diff
@@ -21,7 +27,8 @@ function componentPreview(err, req, res, component, templates) {
       dependencies: _.keys(component.dependencies),
       href: res.conf.baseUrl,
       qs: urlBuilder.queryString(req.query),
-      templates: templates
+      templates,
+      liveReload
```
matteofigus: alphabetic

src/registry/views/component-preview.pug
```diff
@@ -12,3 +12,4 @@ html
       oc.conf.templates = oc.conf.templates || [];
       oc.conf.templates = oc.conf.templates.concat(!{JSON.stringify(templates)});
     script(src=baseUrl+'oc-client/client.js')
+    | !{liveReload}
```
matteofigus: What about when this view is used in the UI? Should we put this here (with a localhost url) only if the registry is on local mode?

src/cli/facade/dev.js
```diff
@@ -43,7 +44,9 @@ module.exports = function(dependencies) {
           if (!hotReloading) {
             logger.warn(strings.messages.cli.HOT_RELOADING_DISABLED);
           } else {
-            cb(components);
+            cb(components, done => {
+              liveReloadServer.refresh('/');
```
matteofigus: how do you pick this url? `/`?

src/registry/views/component-preview.pug
nickbalestra: I'm doing that here -> https://github.com/opentable/oc/pull/635/files#diff-e4f8fa94a02abcf0d568f89a410ab48eR16 (if that view is used on the UI when the registry is not on local mode, nothing is printed)

src/cli/facade/dev.js
nickbalestra: The refresh method need you pass a string for the file that was changed. In our case for the moment that's not relevant so I'm passing some dummy content `/` to simply refresh the listening view (in our case the `preview`). 

(In the future we co-ordinate livereload to work even more closely with out browser-client so that we could perform some sort of hot-reloading. For example immagine using this with the chrome-extention tool to just re-render that specific component...)

nickbalestra:
@matteofigus updated/commented

Issue #379, opened 2017-02-09
Supporting Minio (or other non-AWS S3 solution)


adamgoose:
I'd like to use OC with [minio](https://minio.io). Since it's S3-compliant, it could easily be possible by allowing configuration of the S3 API endpoint (as opposed to the AWS endpoints). Minio also has a requirement of using v4 of the S3 signature algorithm. 

```
var s3  = new AWS.S3({
          accessKeyId: 'YOUR-ACCESSKEYID' ,
          secretAccessKey: 'YOUR-SECRETACCESSKEY' ,
          endpoint: 'http://127.0.0.1:9000' ,
          s3ForcePathStyle: 'true', // needed with minio?
          signatureVersion: 'v4'
                                });
```

http://docs.minio.io/docs/how-to-use-aws-sdk-for-javascript-with-minio-server

adamgoose:
Looks like a conditional update to [these lines](https://github.com/opentable/oc/blob/master/src/registry/domain/s3.js#L17-L22) would do the trick!

matteofigus:
Hi, I made actually a bit of research. OC uses file-based ACLs to secure some files, and unfortunately minio doesn't seem to support that (yet): https://github.com/minio/minio/blob/77a192a7b560f4f7c933280536231996cfb39414/docs/minio-limitations.md#list-of-amazon-s3-object-apis-not-supported-on-minio

I think minio looks like a great project and perhaps it would be possible to ask for that feature. In any case, I would be happy to accept a PR for OC to make additional S3 parameters configurable, such as endpoint, signatureVersion, etc. Would you be able to try?

Only thing, I removed the minio suggestion from the wiki until we ensure it allows full integration - we can add it later.

adamgoose:
Sure thing! Once I dive in a bit deeper, I'll see if I can get a PR made.

matteofigus:
I am closing this at the moment. We can reopen in case we still are interested.

danlaney:
I was just researching this aspect of S3 API support.  We have some fine-grained access control requirements and are looking for open source solutions that might form a back-end for our tools.  We have a draft requirements document [here](https://github.com/DOE-Workflows/SDM-requirements)

kmcrawford:
I have created a PR #662 against this issue.  I created minio in my kubernetes cluster (on GCP) with no external access, so this should minimize security concerns.

[...]

PR #662, opened 2017-09-25
Issue #379 support for minio

kmcrawford:
Add support for minio via endpoint s3 config.

Example:

endpoint: 'http://minio.local:9000',
    path: 'http://minio.local:9000/oc-registry/',
    s3ForcePathStyle: 'true',
    signatureVersion: 'v4',

### DIFF
--- a/src/registry/domain/repository.js
+++ b/src/registry/domain/repository.js
@@ -51,8 +51,8 @@ module.exports = function(conf) {
         const isDir = fs.lstatSync(path.join(conf.path, file)).isDirectory();
         const isValidComponent = isDir
           ? fs
-              .readdirSync(path.join(conf.path, file))
-              .filter(file => file === '_package').length === 1
+            .readdirSync(path.join(conf.path, file))
+            .filter(file => file === '_package').length === 1
           : false;
 
         return isValidComponent;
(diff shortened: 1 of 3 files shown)


kmcrawford:
I am closing this since I misunderstood that nearly everything on s3 is public except server.js.  This complicates things and does add security risks.  Minio only has the ability to mark directories as public/private and not individual files.

matteofigus:
Correct. I don't know minio a lot, but perhaps if you put a cdn on top of it or some sort of proxy, you can add a rule to block server.js to be publicly served?

kmcrawford:
I have figured out how to set file level permissions via minio using bucket policy.  Essentially it will create a resource like below for everything that needs to be public.  The only issue there is a bug with their setBucketPolicy api, where it wipes out the current bucket policy and unable to append.  I spoke to the staff an minio and they said they will fix this next week.  I nearly had this all working locally till I found this bug, so once they fix I can create a new PR that will support this.

I know it's not perfect creating an * on the end of the resource, but since the extension is included it should be unique.

https://github.com/minio/minio-js/issues/618

{
		"Sid": "",
		"Effect": "Allow",
		"Principal": {
			"AWS": ["*"]
		},
		"Action": ["s3:GetObject"],
		"Resource": ["arn:aws:s3:::registry/components/oc-client/0.41.5/template.js*"]
	},

matteofigus:
Fantastic, great job

PR #681, opened 2017-10-02
[DX-185] Refactoring

matteofigus:
This is the first part of a job of cleaning up all the npm stuff. 
The idea is to progressively move all the interactions to npm to an independent module for better modularity and less coupling.

### DIFF
--- a/src/cli/domain/init-template/install-template.js
+++ b/src/cli/domain/init-template/install-template.js
@@ -1,10 +1,9 @@
 'use strict';
 
-const path = require('path');
-const spawn = require('cross-spawn');
-const _ = require('lodash');
+const tryRequire = require('try-require');
 
 const isValidTemplate = require('../../../utils/isValidTemplate');
+const npm = require('../../../utils/npm-utils');
 const strings = require('../../../resources');
 
 module.exports = function installTemplate(options, callback) {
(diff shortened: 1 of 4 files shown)


src/cli/domain/init-template/install-template.js
```diff
@@ -16,29 +15,27 @@ module.exports = function installTemplate(options, callback) {
...
-    stdio: 'inherit'
-  });
+  logger.log(strings.messages.cli.installCompiler(compiler));
 
-  installProc.on('error', () => callback('template type not valid'));
-  installProc.on('close', code => {
-    if (code !== 0) {
-      return callback('template type not valid');
+  npm.installDependency(npmOptions, (err, result) => {
+    const errorMessage = 'template type not valid';
```
nickbalestra: Note: In the future we might want to improve those errors, I found that as we use the same error in different point make it quite hard to debug.

src/cli/domain/init-template/install-template.js
```diff
@@ -16,29 +15,27 @@ module.exports = function installTemplate(options, callback) {
...
-      'node_modules',
-      compiler
-    ));
-
-    if (!isValidTemplate(maybeCompiler, { compiler: true })) {
-      return callback('template type not valid');
+
+    const installedCompiler = tryRequire(result.dest);
+
+    if (!isValidTemplate(compiler, { compiler: true })) {
```
matteofigus: Ah! that should have been `isValidTemplate(installedCompiler...` instead! Bugfix coming in another PR: https://github.com/opentable/oc/pull/694

PR #686, opened 2017-10-03
[DX-185] Removed NPM as dependency

### LINKED ISSUE #593: Move npm to cross-spawn npm
Currently we self-host npm inside the cli to run install commands.

It would be nice to use cross-spawn so that 
1) it could use a separate thread possibly resulting in better performance
2) it would run the command with the npm version that the user has in its machine. This brings simplicity, transparency, and enhances control
3) Consistency: we already use cross-spawn in the templates and other parts - so this would reduce the bundle too

matteofigus:
Closes #593, #622


src/utils/npm-utils.js
```diff
@@ -3,31 +3,66 @@
...
+    args.push('--save-exact');
+    args.push(options.isDev ? '--save-dev' : '--save');
+  }
 
-    const args = ['install'];
+  return args;
+};
+
+const executeCommand = (options, callback) => {
+  const cmd = spawn('npm', options.command, {
```
nickbalestra: Cool here we could hook with yarn support easily in the future

src/utils/npm-utils.js
```diff
@@ -3,31 +3,66 @@
...
+  },
+  installDependency: (options, callback) => {
+    const { dependency, installPath } = options;
+    const npmi = buildInstallCommand(options);
+    const cmdOptions = { installPath, command: [...npmi, dependency] };
 
-    cmd.on('error', () => callback('error'));
-    cmd.on('close', code => {
-      const err = code !== 0 ? code : null;
+    executeCommand(cmdOptions, err =>
```
nickbalestra: That seems like a lot of duplication just to handle a single vs multiple deps install, coudln't we make into a single method?

PR #688, opened 2017-10-04
Decouple s3

kmcrawford:


src/registry/domain/options-sanitiser.js
```diff
@@ -53,11 +53,38 @@ module.exports = function(input) {
...
+
+  if (options.verbosity && options.storage) {
+    options.storage.options.verbosity = options.verbosity;
+  }
+  if (
+    options.storage &&
+    options.storage.adapterType &&
+    !options.storage.adapter
+  ) {
+    switch (options.storage.adapterType) {
```
nickbalestra: I think we could simply check for `options.s3` here without the need for an adapterType. 
In the case that `s3` is not avail, you will need to declaratively pass an adapter and options and rely on the new API. What you think?

src/registry/domain/options-sanitiser.js
```diff
@@ -53,11 +53,38 @@ module.exports = function(input) {
...
+    case 's3':
```
nickbalestra: As mentioned above this will avoid to have a switch statement on the `adapterType`, and just setup a default in case for `s3` config (legacy) overridable with the new API.

src/registry/domain/repository.js
```diff
@@ -10,20 +10,22 @@ const ComponentsCache = require('./components-cache');
...
 const versionHandler = require('./version-handler');
 
 module.exports = function(conf) {
-  const cdn = !conf.local && new S3(conf);
-  const repositorySource = conf.local ? 'local repository' : 's3 cdn';
+  const cdn = !conf.local && new conf.storage.adapter(conf.storage.options);
+  const options = !conf.local && conf.storage.options;
+  const repositorySource = conf.local
+    ? 'local repository'
+    : conf.storage.adapterType + ' cdn';
```
nickbalestra: Perhaps here we could have the adapter exposing his own name instead ?
maybe something like `conf.storage.adapter.type?`

src/registry/domain/repository.js
```diff
@@ -10,20 +10,22 @@ const ComponentsCache = require('./components-cache');
...
-const S3 = require('./s3');
 const settings = require('../../resources/settings');
 const strings = require('../../resources');
 const validator = require('./validators');
 const versionHandler = require('./version-handler');
 
 module.exports = function(conf) {
-  const cdn = !conf.local && new S3(conf);
-  const repositorySource = conf.local ? 'local repository' : 's3 cdn';
+  const cdn = !conf.local && new conf.storage.adapter(conf.storage.options);
```
nickbalestra: Nice I like that now we can just pass the options to its own adapter 👍

src/registry/domain/options-sanitiser.js
kmcrawford: My main reason for using adapterType was the the user would have to specify the following:

adapter: require('oc/src/domain/registry/s3')

It just didn't seem elegant to have people reach down into the source to pull back config info.  With that I thought about removing the adapter from the config, but I also liked how I could add my own if I had a proprietary internal datastore that I didn't want to have in open source land.

src/registry/domain/repository.js
kmcrawford: Funny I had it that way first..., but as stated before I wasn't a fan of 
adapter: require('oc/src/domain/registry/s3')

src/registry/domain/options-sanitiser.js
nickbalestra: Ok, I see what you trying to do. 
So, if we are using `adapterType` mainly for the default adapter (`s3`) to avoid people to reach down to get the default adapter we could just make it optional? if you don't specify an adapter it will use the s3 one that come with oc, otherwise you need to pass your own? This way we don't add an option that is actually only been used for s3.

src/registry/domain/repository.js
nickbalestra: I agree with you and I think we can avoid the user to set `adapter: require('oc/src/domain/registry/s3')` as that will be the default, so you'll only pass an adapter if you want to change that.

PR #689, opened 2017-10-04
ignoring lock files

nickbalestra:


.gitignore
```diff
@@ -21,6 +21,7 @@ results
 
 npm-debug.log
 node_modules
+package-lock.json
```
lobut: aren't `package-lock.json` files intended to be checked in?

nickbalestra: Yes for projects, while there are different opinions for libraries. I think yarn and npm also handle those differ and this is just a story to keep them out until we have a strong opinion/strategy for those. 
This is an interesting read that goes into your direction: https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

PR #694, opened 2017-10-05
Npm install for new module bugfix

matteofigus:
This is a bug introduced in a previous PR.
Look at the link down there for the exact line of code.


src/cli/domain/init-template/install-template.js
```diff
@@ -32,7 +32,7 @@ module.exports = function installTemplate(options, callback) {
 
     const installedCompiler = tryRequire(result.dest);
 
-    if (!isValidTemplate(compiler, { compiler: true })) {
+    if (!isValidTemplate(installedCompiler, { compiler: true })) {
```
nickbalestra: 👍 nice catch

PR #659, opened 2017-09-22
Issue #464: Replaces Grunt with NPM scripts

jakeclements:
Fixes #464 
- Replaces Grunt with NPM scripts
- Removes text-limit on changelog messages
- Updates contributing and readme docs with run-script commands

PR #704, opened 2017-10-09
[DX-202] Execute all the tests

matteofigus:
Fixes a bug introduced on #659 - the promise was resolved on the first file, not making all the tests to be correctly executed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ test/fixtures/test.tar.gz
 test/fixtures/targz-test
 
 logintervals.md
+
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ src/components/base-component-jade/_package
 
 test/fixtures/test.tar.gz
 test/fixtures/targz-test
+
+logintervals.md
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
   "scripts": {
     "build": "node tasks/build.js",
     "git-stage": "node tasks/git-stage.js",
-    "major": "npm run version-major && npn run test-silent && npm run git-stage",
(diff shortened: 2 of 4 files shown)


matteofigus:
So, there are many other bugs. I think this PR attempts to fix the testing part and the build part (build needs to happen before the tests).

The Changelog generation connected to the version tasks seems not to work and I'll follow up with a fix in another PR

.gitignore
```diff
@@ -29,3 +29,5 @@ src/components/base-component-jade/_package
 
 test/fixtures/test.tar.gz
 test/fixtures/targz-test
+
+logintervals.md
```
mattiaerre: add `newline` please,

package.json
```diff
@@ -7,16 +7,17 @@
...
-    "build": "node tasks/build.js && npm run test",
-    "changelog": "node tasks/changelog.js",
+    "build": "node tasks/build.js",
     "git-stage": "node tasks/git-stage.js",
-    "major": "node tasks/version.js --type=\"major\"",
-    "minor": "node tasks/version.js --type=\"minor\"",
-    "patch": "node tasks/version.js --type=\"patch\"",
+    "major": "npm run version-major && npm run test-silent && npm run git-stage",
+    "minor": "npm run version-minor && npm run test-silent && npm run git-stage",
+    "patch": "npm run version-patch && npm run test-silent && npm run git-stage",
```
mattiaerre: you can also use the native `npm version <update_type>` to update version on npm btw

matteofigus: This is a bit different as it updates the component's version too.

.gitignore
matteofigus: done

PR #710, opened 2017-10-11
[WIP][DX-185] install compiler inside each components' dir 

matteofigus:
This fixes https://github.com/opencomponents/oc-template-react/issues/18


src/cli/domain/handle-dependencies/require-template.js
```diff
@@ -0,0 +1,51 @@
...
+    template
+  );
+
+  [
+    componentRelativePath,
+    template,
+    localTemplate,
+    relativeTemplate
+  ].forEach(pathToTry => {
+    ocTemplate = ocTemplate || cleanRequire(pathToTry, { justTry: true });
```
nickbalestra: why not using `some` instead of forEach so that you breakout of the loop as soon it get required? Am I missing something?

src/utils/clean-require.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const tryRequire = require('try-require');
+const _ = require('lodash');
```
nickbalestra: I guess no need for lodash here?

src/cli/domain/handle-dependencies/require-template.js
matteofigus: I don't know about some. Perhaps let's do it in another PR as optimisation?

src/utils/clean-require.js
matteofigus: done

PR #722, opened 2017-10-23
[DX-218] Expose cli to be used programmatically

### LINKED ISSUE #441: Running local instance of registry via node does not create _package folders
**Background** : We are working on enabling developers to be able to write integration test suites for their components, to do this one of the things needed is to be able to bring up and stop a local instance of the registry.

**Code** : 
```
startRegistry: function(done) {
            registry = new oc.Registry({
                baseUrl: 'http://localhost:3030/',
                local: true,
                port: 3030,
                hotReloading: false,
                path: path.resolve('../'),
                env: {name: 'local'},
                verbosity: 1,
                discovery: true
            });
            registry.start(done);
        }
```
Pretty straight forward snippet. Note that we need discovery true so that we can use phantomJS or nightmare js or casper ... to be able to load the component in preview mode.

**Problem** : So the registry did start up but it does not generate the _package folder for the component, which means we cannot really load the component in preview mode.

Thoughts?

nickbalestra:
Fixes #441 by exposing the dev interface and allowing acceptance testing in a cleaner way

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,8 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
-const Local = require('./cli/domain/local');
-
 module.exports.cli = require('./cli/programmatic-api');
 module.exports.Client = require('oc-client');
 module.exports.Registry = require('./registry');
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,18 @@
 'use strict';
 
+const _ = require('lodash');
+const Local = require('./cli/domain/local');
+
+function cliCommand(cmd, options, cb) {
+  const dependencies = {
+    local: new Local(),
+    logger: options.logger || { log() {} }
+  };
+  const opts = _.omit(options, 'logger');
+
+  require(`./cli/facade/${cmd}.js`)(dependencies)(opts, cb);
+}
+
+module.exports.cli = cliCommand;
 module.exports.Client = require('oc-client');
 module.exports.Registry = require('./registry');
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -1,4 +1,18 @@
...
+  const dependencies = {
+    local: new Local(),
+    logger: options.logger || { log() {} }
+  };
+  const opts = _.omit(options, 'logger');
+
+  require(`./cli/facade/${cmd}.js`)(dependencies)(opts, cb);
+}
+
+module.exports.cli = cliCommand;
```
matteofigus: can we move lines 3-14 to own file inside the cli folder?

src/index.js
```diff
@@ -1,4 +1,18 @@
 'use strict';
 
+const _ = require('lodash');
+const Local = require('./cli/domain/local');
+
+function cliCommand(cmd, options, cb) {
+  const dependencies = {
+    local: new Local(),
+    logger: options.logger || { log() {} }
```
matteofigus: the logger needs more methods: https://github.com/opentable/oc/blob/master/src/cli/logger.js#L14-L17

src/index.js
```diff
@@ -1,4 +1,8 @@
 'use strict';
 
+const _ = require('lodash');
+const Local = require('./cli/domain/local');
```
matteofigus: forgot about those 2 ^ ? Everything else LGTM

nickbalestra:
@matteofigus 
Fixed, there also were some wrong paths, and I've also added some unit tests. Should be good to review now

PR #741, opened 2017-11-07
Don't create package locks on install

nickbalestra:


nickbalestra:
I see that you actually doing that for all the commands to be run, so no need for this.

PR #763, opened 2017-11-21
passing prefix option to dev cmd

debelbot:
The prefix option would be very useful to us while developing components and testing them locally in different contexts


matteofigus:
Thanks for your help @debelbot

PR #768, opened 2017-11-27
[DX-247] dev/publish cli mode - prevent some deps to run in an infinite loop

### LINKED ISSUE #730: oc-client 0.41.15 does not properly compile some dependencies
Today I've found our CI environment failing on OC deploy, due to the following error:

imatge

We didn't change anything from the previous working deploy to this, but the oc-client version changed. Using oc-client 0.41.13 works as expected, but the current doesn't allow us to properly publish to the registry.

matteofigus:
Fixes #730

### DIFF
--- a/src/cli/domain/handle-dependencies/get-missing-dependencies.js
+++ b/src/cli/domain/handle-dependencies/get-missing-dependencies.js
@@ -9,7 +9,7 @@ module.exports = dependencies => {
   const missing = [];
   _.each(dependencies, (version, dependency) => {
     const pathToModule = path.resolve('node_modules/', dependency);
-    if (!cleanRequire(pathToModule, { justTry: true })) {
+    if (!cleanRequire(pathToModule, { justTry: true, resolve: true })) {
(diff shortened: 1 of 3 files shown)


elboletaire:
Sorry to ask this twice, but apparently my first comment went unnoticed. Could it be possible to add some cli arg to avoid checking for dependencies? This is something we already do always before pushing to the registry, so oc makes it redundant and tries to do something that has been recently done. In our side would have much sense to be able to disable this.

/cc @nickbalestra @matteofigus

nickbalestra:
I wouldn't be against a flag to skip packaging for the `publish`. Will you be up to attempt a PR for this? Happy to help if you need

elboletaire:
Yeah ofc @nickbalestra. For the moment I'll open it, and if I have time I'll try to do something, but the main problem is the time, you know... 😅

nickbalestra:
Awesome. Time...I know 😅 , no rush and keep me posted

elboletaire:
btw created #769 as a follow-up

Issue #789, opened 2017-12-15
Error Message from repository.getComponentInfo logged with [Object object]


bassettsj:
## *Who* is the bug affecting?
Registry maintainers

## *What* is affected by this bug?
The error message is not very helpful.

## *When* does this occur?
1. when `repository.geComponentInfo`'s callback is called with a non error object from a storage error it does not give a helpful message
2. observed in the preview window:
`"component not available: [object Object]"`

## *Where* on the platform does it happen?
api

## *How* do we replicate the issue?

https://demo-oc-registry.herokuapp.com/tommys-component/1.0.1/~preview

## Expected behavior (i.e. solution)
Expected to give the actual error message which is:

`{ code: 'file_not_valid',
  msg: 'File "components/tommys-component/1.0.0/package.json" not valid' }`

## What version of OC, Node.js and OS are you using?
```
{ http_parser: '2.7.0',
  node: '9.3.0',
  v8: '6.2.414.46-node.15',
  uv: '1.18.0',
  zlib: '1.2.11',
  ares: '1.13.0',
  modules: '59',
  nghttp2: '1.25.0',
  openssl: '1.0.2n',
  icu: '60.1',
  unicode: '10.0',
  cldr: '32.0',
  tz: '2017c' }

oc@0.42.12

OSX 10.13.1
arch: 'x64',
platform: 'darwin',
```

## Other Comments

matteofigus:
Thanks for the issue @bassettsj - I am investigating.

bassettsj:
@matteofigus I added a failing unit test showing the error. I will open the PR, sorry if this duplicating work.

matteofigus:
No problem at all @bassettsj I was looking at the code but didn't write any code yet. If you would want to attempt a PR for the tests and a fix, that would super fantastic, want to give it a go?

catt231:
Hi I've been trying to use the starter kit to create my first registry but I'm only seeing the JSON not valid error above - could you point me in any direction of what's going wrong?

See example oc-client error which occurs on standard example of starter kit deployment

https://starter-kit-itrqyqhfdt.now.sh/oc-client/0.42.16/~info

matteofigus:
@catt231 we managed to reproduce and now it should be fixed by #811 and available on the latest version (0.42.18) - can you try to upgrade OC and retry?

catt231:
fixed thanks!

PR #792, opened 2017-12-15
Add test for issue #789 showing error message [Object object]

bassettsj:
Please do not create a Pull Request without creating an issue first.
Any change needs to be discussed before proceeding.
Failure to do so may result in the rejection of the pull request.

Demonstrating the bad error message I was seeing after setting up demo registry on heroku. This is showing the problematic error message I saw.

**Test plan (required)**

Running unit tests should prove if the error is fixed or not.

**Closing issues**

#789  would be closed if I fixed the failing test as well, but I would like to discuss if it is an actual bad error condition or not.

### DIFF
--- a/src/registry/domain/repository.js
+++ b/src/registry/domain/repository.js
@@ -14,6 +14,7 @@ const settings = require('../../resources/settings');
 const strings = require('../../resources');
 const validator = require('./validators');
 const versionHandler = require('./version-handler');
+const errorToString = require('../../utils/error-to-string');
 
 module.exports = function(conf) {
   const cdn = !conf.local && new conf.storage.adapter(conf.storage.options);
@@ -161,7 +162,10 @@ module.exports = function(conf) {
           version,
           (err, component) => {
             if (err) {
-              return callback(`component not available: ${err}`, null);
+              return callback(
+                `component not available: ${errorToString(err)}`,
+                null
+              );
             }
             callback(null, _.extend(component, { allVersions }));
           }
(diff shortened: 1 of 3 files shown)


matteofigus:
I think this sounds legit. When we have an error from there, we should show something meaningful in the api response.

src/utils/error-to-string.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function errorToString(err) {
```
bassettsj: @matteofigus I wasn't sure where to put this but but it would just see if it has a message attribute that is a string and returns that.

src/registry/domain/repository.js
```diff
@@ -258,12 +262,9 @@ module.exports = function(conf) {
       )}`,
 
     getStaticFilePath: (componentName, componentVersion, filePath) =>
-      `${repository.getComponentPath(
-        componentName,
-        componentVersion
-      )}${conf.local
-        ? settings.registry.localStaticRedirectorPath
-        : ''}${filePath}`,
+      `${repository.getComponentPath(componentName, componentVersion)}${
```
bassettsj: Didn't change this, but it seems prettier seems to want to change it.

matteofigus: No worries, that is by design. Perhaps we should mention that in the contributing guidelines to avoid confusion :)

matteofigus:
Closes #789

src/utils/error-to-string.js
```diff
@@ -0,0 +1,11 @@
...
+  const isString = typeof error === 'string';
```
matteofigus: can we use lodash's _.isString? This is for consistency with other bits in the codebase

src/utils/error-to-string.js
```diff
@@ -0,0 +1,11 @@
...
+  const hasMessage = err != null && typeof err.msg === 'string';
+  if (isString) {
+    return err;
+  } else if (hasMessage) {
+    return err.msg;
+  }
+
+  return err + '';
```
matteofigus: what about `err.toString()` which perhaps is a bit more readable

bassettsj: Hmm... That might cause errors like `TypeError: Cannot read property 'toString' of undefined` if it is null or undefined.

src/utils/error-to-string.js
bassettsj: changed

src/utils/error-to-string.js
matteofigus: good point!

PR #794, opened 2017-12-20
[DX-296] Mocks can have the same signature as real plugins

matteofigus:
This allows a plugin to have, apart from the simplified signature (exporting a synchronous execution function) a regular signature (similar to the real plugins) with an exported asynchronous register() function and a synchronous exported execute() function.

This is part of a bigger work to allow mocks to be more usable and possibly wrapping the whole real plugin - for better local DX and testing.

PS
In case of plugin initialisation error, I improved the error message providing some code snippets for modelling a mock with the correct signature:
screen shot 2017-12-20 at 15 26 51

### DIFF
--- a/src/cli/domain/get-mocked-plugins.js
+++ b/src/cli/domain/get-mocked-plugins.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const colors = require('colors/safe');
 const fs = require('fs-extra');
 const path = require('path');
 const _ = require('lodash');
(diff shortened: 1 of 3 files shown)


src/cli/domain/get-mocked-plugins.js
```diff
@@ -1,73 +1,77 @@
...
       return undefined;
     } else {
-      const getParent = function(x) {
-          return x
-            .split('/')
-            .slice(0, -1)
-            .join('/');
-        },
-        parentDir = pathToResolve ? getParent(pathToResolve) : rootDir;
+      const getParent = x =>
```
nickbalestra: I think is never too bad to name things for what they are: `pathToResolve` or similar in this case instead of `x`

PR #795, opened 2017-12-21
 [DX-324] full async plugins initialisation

matteofigus:
The reason here is an old design decision. I think that given the way plugins evolved, it makes much more sense to wait express to start after all the plugins are fully initialised.

### DIFF
--- a/src/registry/domain/plugins-initialiser.js
+++ b/src/registry/domain/plugins-initialiser.js
@@ -81,8 +81,10 @@ module.exports.init = function(pluginsToRegister, callback) {
   };
 
   const loadPlugin = function(plugin, cb) {
+    const done = _.once(cb);
+
     if (registered[plugin.name]) {
-      return cb();
+      return done();
     }
 
     if (!plugin.register.dependencies) {
@@ -90,7 +92,7 @@ module.exports.init = function(pluginsToRegister, callback) {
     }
 
     if (!dependenciesRegistered(plugin.register.dependencies)) {
-      return defer(plugin, cb);
+      return defer(plugin, done);
     }
 
     const dependencies = _.pick(registered, plugin.register.dependencies);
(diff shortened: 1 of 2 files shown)


src/registry/domain/plugins-initialiser.js
```diff
@@ -81,27 +81,28 @@ module.exports.init = function(pluginsToRegister, callback) {
   };
 
   const loadPlugin = function(plugin, cb) {
+    const done = _.once(cb);
```
matteofigus: This allows robustness against plugins that would erroneously call the done() callback multiple times:
https://github.com/opencomponents/oc-hobknob/blob/master/index.js#L25

src/registry/domain/plugins-initialiser.js
```diff
@@ -81,27 +81,28 @@ module.exports.init = function(pluginsToRegister, callback) {
...
-      dependencies,
-      plugin.callback || _.noop
-    );
-    registered[plugin.name] = plugin.register.execute;
-    cb();
+    plugin.register.register(plugin.options || {}, dependencies, err => {
+      const pluginCallback = plugin.callback || _.noop;
+      pluginCallback(err);
+      registered[plugin.name] = plugin.register.execute;
+      done(err);
```
matteofigus: now the callback is called actually after the registration is completed

test/unit/registry-domain-plugins-initialiser.js
```diff
@@ -105,29 +105,24 @@ describe('registry : domain : plugins-initialiser', () => {
...
           name: 'isFlagged',
           register: {
-            register: function(options, deps, cb) {
-              flag = true;
-              cb();
+            register: (options, deps, cb) => {
+              setTimeout(() => {
+                flag = true;
+                cb();
+              }, 10);
```
matteofigus: With this artificial timeout, I am testing that the callback is actually called after the registration (test failed before the code change)

PR #813, opened 2018-01-25
[DX-366] Allow usage of protocol  in storage.options.path

nickbalestra:
If no protocol is provided (current legacy usage) default to https.
cc @johnparn

related issue - https://github.com/opencomponents/storage-adapters/issues/34

📝  Update [documentation](https://github.com/opentable/oc/wiki/Registry#registry-configuration) once this get published

### DIFF
--- a/src/registry/domain/repository.js
+++ b/src/registry/domain/repository.js
@@ -19,6 +19,10 @@ const errorToString = require('../../utils/error-to-string');
 module.exports = function(conf) {
   const cdn = !conf.local && new conf.storage.adapter(conf.storage.options);
   const options = !conf.local && conf.storage.options;
+  const storagePath =
+    options && options.path.match(/^http/)
+      ? options.path
+      : `https:${options.path}`;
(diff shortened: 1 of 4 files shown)


src/registry/domain/repository.js
```diff
@@ -19,6 +19,10 @@ const errorToString = require('../../utils/error-to-string');
 module.exports = function(conf) {
   const cdn = !conf.local && new conf.storage.adapter(conf.storage.options);
   const options = !conf.local && conf.storage.options;
+  const storagePath =
```
matteofigus: can we rather bring this logic here: https://github.com/opentable/oc/blob/master/src/registry/domain/options-sanitiser.js

nickbalestra:
Thanks for the suggestion @matteofigus , totally agree. 
I've also discovered we were missing some tests for all the sanitization we do related to storage adapters/legacy cases so I've added coverage for all the sanitization we do there, plus one for the new path

PR #872, opened 2018-05-04
[DX-497] OC Clean

matteofigus:
This introduces an handy oc clean command, pretty useful with monorepos of OCs that now contain templates and subdeps. The functionality is pretty similar to lerna's clean.

### DIFF
--- a/src/cli/commands.js
+++ b/src/cli/commands.js
@@ -3,6 +3,20 @@
 module.exports = {
   usage: 'Usage: $0 <command> [options]',
   commands: {
+    clean: {
+      cmd: 'clean <dirPath>',
+      example: { cmd: '$0 clean components --yes' },
+      description: `Remove the node_modules directory from each component's subfolder`,
+      options: {
+        yes: {
+          boolean: true,
+          description: 'Skip all confirmation prompts',
+          default: false
+        }
+      },
+      usage: 'Usage: $0 clean <dirPath>'
+    },
+
     dev: {
       cmd: 'dev <dirPath> [port] [baseUrl]',
       example: {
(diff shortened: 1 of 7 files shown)


src/resources/index.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
-const colors = require('colors/safe');
+const { green, yellow } = require('colors/safe');
+
```
matteofigus: I moved things around in this file to optimise readability, definitely doing another separate PR afterwards to reorganise this stuff

src/cli/commands.js
```diff
@@ -3,14 +3,28 @@
...
+
     dev: {
       cmd: 'dev <dirPath> [port] [baseUrl]',
       example: {
         cmd:
           '$0 dev ../all-components 3001 127.0.0.1:3001 --fallbackRegistryUrl=http://anotherhost:anotherport/'
       },
       description:
-        'Runs a local oc test registry in order to develop and test components',
+        'Run a local oc test registry in order to develop and test components',
```
matteofigus: Took inspiration from lerna here, all the actions are intended as 2nd person rather than 3rd person intending the action is done by the user not the CLI really :)

src/cli/commands.js
```diff
@@ -43,7 +57,7 @@ module.exports = {
           default: ''
         }
       },
-      usage: 'Usage: $0 dev <dirName> [port] [baseUrl] [options]'
+      usage: 'Usage: $0 dev <dirPath> [port] [baseUrl] [options]'
```
matteofigus: I updated this example to reflect the naming we use everywhere else

PR #911, opened 2018-06-30
add component name to context

kmcrawford:
Add component name to context obj.  Since we have dynamic naming of branch components I have no way of knowing what the name of the current component I am in.  Which is needed for a configuration plugin I am working on.


matteofigus:
@kmcrawford thanks for the PR.

The way we usually handle this is by using its own package.json on build time
`const name = require('./package.json').name;` which *should* be a bit better to reduce runtime computation.

Would that work for you too?

kmcrawford:
Sounds good. 👍🏻

PR #931, opened 2018-09-06
expose express app to registry

kmcrawford:
Expose express app to the registry to allow middleware configurations.

### DIFF
--- a/src/registry/index.js
+++ b/src/registry/index.js
@@ -104,6 +104,7 @@ module.exports = function(options) {
     close,
     on: eventsHandler.on,
     register,
-    start
+    start,
+    app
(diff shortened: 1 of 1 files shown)


nickbalestra:
Hey Ken, thanks for the PR!
Have you tried instantiate the registry by `const registry = oc.Registry(config)`.
You can then access the express app by `registry.app` which then allows you to attach any middleware to express and customise as you prefer.

Here is where it happen in the code -> https://github.com/opencomponents/oc/blob/master/src/registry/index.js#L92

seif:
Thanks @kmcrawford @nickbalestra

PR #941, opened 2018-10-01
[Suggestion] Handling relative path when initiating component

jankowiakmaria:
Not creating an issue - just a small PR that allows using relative path when initiating new component.

**Why this change?**

In our team all the components are in `/src/components` directory. All the grunt/npm tasks are defined at the top level though. When we want to create a new component it requires us to first go two levels down to do the `init` and then go back up to run registry locally or run the tests.

Supporting relative path would allow us to slightly reduce number of steps during the development process.

This change affects only init CLI.

**Test plan (required)**

Output when running `node src/cli/ init test-components/hello-test`
image

### DIFF
--- a/src/cli/commands.js
+++ b/src/cli/commands.js
@@ -61,13 +61,13 @@ module.exports = {
     },
 
     init: {
-      cmd: 'init <componentName> [templateType]',
+      cmd: 'init <componentPath> [templateType]',
       example: {
         cmd: '$0 init test-component oc-template-es6'
       },
       description:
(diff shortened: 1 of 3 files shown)


matteofigus:
Thanks @jankowiakmaria can I ask you to update the github wiki? Thanks!

jankowiakmaria:
Thanks for merging - will have a look.

jankowiakmaria:
@matteofigus https://github.com/opencomponents/oc/wiki/Cli#init (updated + added example)

PR #1024, opened 2019-04-16
set headers for getComponents

### LINKED ISSUE #1023: context.setHeader from server.js not being applied when calling getData function oc-react-template
## This is effecting anyone one that is using a client getData call from oc-react-template and the server does a context.setHeader the header is not being set.

What I found is that headers are being set in component.js (below), but when a getData call is being make it is using components.js 

Setting headers:
https://github.com/opencomponents/oc/blob/master/src/registry/routes/component.js#L29

Not setting headers:
https://github.com/opencomponents/oc/blob/master/src/registry/routes/components.js#L64

I have fixed this locally by adding this code below:
```
const setHeaders = (results, res) => results.forEach(result => _.isEmpty(result.headers)|| res.set(result.headers));
```
In the result calling the function above:
```
(err, results) => {setHeaders(results,res); res.status(200).json(results);}
```

kmcrawford:
Fix issue: 
https://github.com/opencomponents/oc/issues/1023

**Test plan (required)**
Setting a header using context in server.js
image

From: client side:
image

Result: No cookie..

After change: (Cookie!)
image

**Closing issues**
closes #1023

### DIFF
--- a/src/registry/routes/components.js
+++ b/src/registry/routes/components.js
@@ -9,6 +9,7 @@ const strings = require('../../resources');
 
 module.exports = function(conf, repository) {
   const getComponent = new GetComponentHelper(conf, repository);
+  const setHeaders = (results, res) => results.forEach(result => _.isEmpty(result.headers) || res.set(result.headers));
 
   return function(req, res) {
     const components = req.body.components,
@@ -61,7 +62,10 @@ module.exports = function(conf, repository) {
           result => callback(null, result)
         );
       },
-      (err, results) => res.status(200).json(results)
+      (err, results) => {
+        setHeaders(results,res); 
+        res.status(200).json(results);
+      }
     );
   };
 };
--- a/src/registry/routes/components.js
+++ b/src/registry/routes/components.js
@@ -9,7 +9,12 @@ const strings = require('../../resources');
 
 module.exports = function(conf, repository) {
   const getComponent = new GetComponentHelper(conf, repository);
-  const setHeaders = (results, res) => results.forEach(result => _.isEmpty(result.headers) || res.set(result.headers));
+  const setHeaders = (results, res) => {
+    if (!results || results.length !== 1 || !results[0] || !res.set) {
+      return;
+    }
+    res.set(results[0].headers);
+  };
 
   return function(req, res) {
     const components = req.body.components,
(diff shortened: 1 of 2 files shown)


kmcrawford:
I noticed the tests failed, but I believe that these tests were passing due to the missing code.  There is no logic to prevent headers from being added.

matteofigus:
By design we don't set the headers on the POST endpoint, being that originally intended to be used as batch. That is to keep the API simple, in fact what would happen if 2 requested components set 2 different values for the same header? So at the moment, we send the headers back in the response and leave the client to do the required actions.

I admit that in the react template we use the the batch endpoint mostly for the comfort of being able to POST rather than being interested on multiple components (as the request is only for a component), so I think this is the reason we are facing this issue now. We should have though of that before, and I apology for that.

In my opinion, we have some ways to solve this:
1) One is to create a POST route for a single component, exactly as we have the GET, that is able to set the headers. Then to move the react template to use that.
2) The second, is to use the current POST batch endpoint and to find a way to set the headers.
  * (2a) One possibility would be to parse all the union of the set headers, and to set only the unique values, which would be a small variant of this PR. I am still unsatisfied about the fact conflicts may be solved by component order, which would be slightly opaque as API (for instance, component 1 has cache 1 minute, component 2 has cache never => cache is 1 minute of everything - if flipping the order => cache is never) - In both scenarios, I feel like the result is less then ideal as it could break a contract between component creators and component consumers, and this goes against the OC philosophy
  * (2b) Another one is to maybe set the headers ONLY if the requested components number is one. That would solve our current problem, is not quite elegant, but somehow doesn't enter the territory of betting on solving the conflicts. Somehow, it tries to be smart on the easiest scenario, but doesn't go too far with that.
* (2c) The last one, is to really make a smart engine to parse different kinds of headers and setting them only in case of no conflicts - unfortunately this is hard, slightly better than (2b), but still has no solution in case of conflicts.

Generally I like both 1 and 2b. I think they are equally simple to implement, but (1) needs changes to the react template too, and all the components using that template, so perhaps (2b) would be a decent compromise both as end solution of the problem, or as a first step.

What do you think @kmcrawford? Do you agree with my assumptions? Do you have other ideas?

kmcrawford:
@matteofigus I am in favor of 2b. Let me take a look at updating this solution.

matteofigus:
Thanks a lot @kmcrawford 
I just published this as breaking (0.47.0) as this could potentially break someone's expectations.

PR #1056, opened 2019-06-12
Added the support for changing the post request payload size

### LINKED ISSUE #1054: Adding support for large request size up till 10 mb(now gives error payloadtoolarge )
We are passing large answers as a post request to open component and if it exceeds 100KB, request gets rejected

Node (express ) rejects the request and throws error as payload too large

1. Just Create any open component and make a post request to the open component with post body size exceeding 100 KB

It should not have rejected the request and should have given the request to open component code

## Other Comments

Suggested solution:
We are currently using tag 0.42.16 and On line no 24 and 25, We need to add the parameter 'limit' which puts limit on size of request body
https://github.com/opencomponents/oc/blob/v0.42.16/src/registry/middleware/index.js#L24
So that it look likes below:
app.use(bodyParser.json({limit: '10mb', inflate: true}));
app.use(bodyParser.urlencoded({limit: '10mb', extended: true}));

abhinavbansal19961996:
I want to resolve #1054 by adding support for Post request payload size parameter which we can send through registry.
Earlier, we were able to make post request with body size max up to 100KB. (Default: https://github.com/expressjs/body-parser See limit section)

Now, in this PR I have made the size to be configurable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "oc",
-  "version": "0.48.0",
+  "version": "0.48.4",
   "description": "A framework for developing and distributing html components",
   "main": "./src/index.js",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "oc",
-  "version": "0.48.4",
+  "version": "0.48.0",
   "description": "A framework for developing and distributing html components",
   "main": "./src/index.js",
   "bin": {
--- a/src/components/oc-client/package.json
+++ b/src/components/oc-client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "oc-client",
   "description": "The OpenComponents client-side javascript client",
-  "version": "0.48.0",
+  "version": "0.48.4",
   "repository": "https://github.com/opencomponents/oc/tree/master/components/oc-client",
   "author": "Matteo Figus <matteofigus@gmail.com>",
   "oc": {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oc",
-  "version": "0.48.0",
+  "version": "0.48.4",
```
matteofigus: Can you please revert this? The version will be bumped during the publish

src/components/oc-client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "oc-client",
   "description": "The OpenComponents client-side javascript client",
-  "version": "0.48.0",
+  "version": "0.48.4",
```
matteofigus: Can you please revert this? The version will be bumped during the publish

package.json
abhinavbansal19961996: Thanks @matteofigus  I have updated it. Can you please check?

abhinavbansal19961996: Thanks @matteofigus  I have updated it. Can you please check?

matteofigus:
Just published the new version. Will update the docs now.

abhinavbansal19961996:
> Just published the new version. Will update the docs now.

@matteofigus I checked the doc https://github.com/opencomponents/oc/wiki/Registry.
postRequestPayloadSize: "Sets the limit for the post body size. Example: 50mb. When unset, there is no limit"
^^ I think it needs correction. When unset, there is a default limit of 100KB. (https://github.com/expressjs/body-parser)

Issue #1122, opened 2019-10-10
Node 12 version upgrade


julidegulen:
Node 12 step should be added to Travis CI. 
Some tests might be failing should fixed because expected and returned error messages are not "equal". error messages "contains" expected error messages.

PR #1123, opened 2019-10-10
update travis.yml for node version 12

julidegulen:
Fixes issue  #1122
Update node 12 version for travis CI.

Some tests are failing in Node 12 because a module `oc-template-react` is not available for node 12 and despite the error and expected messages are the same for missing module case; `catch(err)` fails because of comparison ` to.equal( )` method. Another failing test again uses `to.equal( )`  method and to compare expected message and error message.  `to.equal( )` method is replace with  `==` 

Update node version for travis CI

**Test plan (required)**

Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.

**Closing issues**

Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes (if such).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ matrix:
     - node_js: "8"
     - node_js: "9"
     - node_js: "10"
+    - node_js: "12"
 
 notifications:
   email: false
--- a/test/acceptance/registry.js
+++ b/test/acceptance/registry.js
@@ -410,8 +410,10 @@ describe('registry', () => {
     });
 
     it('should respond with rendering error', () => {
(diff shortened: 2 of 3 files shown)


julidegulen:
Some tests in Node 12 fails because a module `oc-template-react` is not available for node 12 and despite the error and expected messages are the same for missing module case; `catch(err)` fails because of comparison ` to.equal( )` method. Another failing test again uses ` to.equal( )` method and to compare expected message and error message.  to.equal( )` method is replace with 
`==`

matteofigus:
Great! Thanks for your help :)

PR #1180, opened 2021-02-18
Respect cdn.maxConcurrentRequests in components-list

### LINKED ISSUE #1179: component-list doesn't respect maxConcurrentRequests
component-list doesn't respect `maxConcurrentRequests` defined in storage adapter.

component creators, registry mainteniers

oc registry

## *When* does this occur?
when publishing a component to registry. More specifically - when refreshing component cache.

## *Where* on the platform does it happen?
oc registry

1. Set up oc registry using [oc-azure-storage-adapter](https://www.npmjs.com/package/oc-azure-storage-adapter)
2. Publish a lot of components (more than 20, it was 60 in my case)
3. When publishing a new component, there are chances that it will fail with ECONNRESET error. That's because the adapter defines `maxConcurrentRequests: 20`, but when oc refreshes component list, it doesn't take `maxConcurrentRequests` into account.

Number of concurrent connections on cache refresh is limited to `maxConcurrentRequests`.

oc: 0.48.0, node: v14.8.0, os: Windows, Linux

qbikez:
Storage adapter defines `maxConcurrentRequests`, which was used in `components-details`, but not in `components-list`. This causes ECONNRESET errors in azure storage adapter, because too many concurrent connections are made.
Fix by using async.mapLimit.

**Closing issues**

closes #1179

### DIFF
--- a/src/registry/domain/components-cache/components-list.js
+++ b/src/registry/domain/components-cache/components-list.js
@@ -42,7 +42,7 @@ module.exports = (conf, cdn) => {
             return callback(err);
           }
 
-          async.map(components, getVersionsForComponent, (errors, versions) => {
+          async.mapLimit(components, cdn.maxConcurrentRequests, getVersionsForComponent, (errors, versions) => {
             if (errors) {
               return callback(errors);
             }
--- a/test/unit/registry-domain-components-cache.js
+++ b/test/unit/registry-domain-components-cache.js
@@ -8,7 +8,8 @@ describe('registry : domain : components-cache', () => {
   const mockedCdn = {
     getJson: sinon.stub(),
     listSubDirectories: sinon.stub(),
-    putFileContent: sinon.stub()
+    putFileContent: sinon.stub(),
+    maxConcurrentRequests: 10
   };
 
   const baseOptions = {


kmcrawford:
What happens when the other non-experimental storage-adapters don't define maxConcurrentRequests?

pbazydlo:
it's already used in `component-details`, so this property should be defined in all storage adapters, or it would be broken already I think? https://github.com/opencomponents/oc/blob/a4da36da9da78e9307c3fdb85302422e3a8959f1/src/registry/domain/components-details.js#L36

pbazydlo:
it seems other adapters also define this property: https://github.com/opencomponents/storage-adapters/search?q=maxConcurrentRequests

pbazydlo:
@qbikez there seem to be some failing tests, can you check those?

kmcrawford:
It's been a while since I've looked at the code. I would have expected that to be a config in to storage object. My mistake.

matteofigus (approved):

matteofigus:
LGTM - feel free to merge

pbazydlo:
cheers :) can you publish new version @matteofigus? I can't remember if there are any steps apart from `npm install;npm publish` :P

matteofigus:
ok np published as `v0.48.16`

PR #1181, opened 2021-02-26
Handle scoped packages when removing version

Taiters:

Fix for https://github.com/opencomponents/oc/issues/1178 (Scoped packages being incorrectly handled when removing the version)

When stripping the version from a package string, we first separate the base name and directory. This is done so we can separate the scope (if present) from the package name, before using `split` to remove the package's version. Once removed, we return the full package with `path.join`

I've moved this code into a `split-version` utility, which is now called by `link-missing-dependencies` and `npm-utils`.

**Test plan (required)**

Unit tests have been added for `split-version.js` to cover the following cases:

* Passing a non-scoped package, both with and without a version
* Passing a scoped package, both with and without a version


matteofigus (approved):

matteofigus:
Thanks!

Issue #769, opened 2017-11-28
Add cli argument to skip dependencies packaging


elboletaire:
In some scenarios people tend to compile dependencies by their own before doing `oc publish`, so let's add a cli argument, something like `--skip-dependencies` or `--skip-build`, to allow us skip the dependencies installation later again via oc.

elboletaire:
If I start working on it I'd assign it to me. Until then, I leave it open for others to collaborate.

matteofigus:
So, given we have an `oc package` command (which contains the `_package` folder, perhaps I would call this `--skip-package` - so it would just publish whatever is inside `_package` (skipping the dependencies' check and all of that. What do you think?

elboletaire:
I don't know the internals like you, so I guess your argument will probably fit better than mine :_D

matteofigus:
@elboletaire look at @NimaSoroush's PR that we just merged. Sounds like very similar but for packaging (skips install and does symlinking) https://github.com/opencomponents/oc/pull/971

elboletaire:
Thanks @matteofigus. Thing is I wasn't able to dedicate more time to this since last time, but I'll look into it as soon as I've time again :smiley:

ricardo-devis-agullo:
I was thinking on taking on this issue and I think there's 3 use cases to consider when doing this new command.

## 1. Mirroring what publish does on the same component
`oc package your-component-folder`
`oc publish your-component-folder --skip-package`

In this case the scenario is you have your component, you package first, it creates the `_package` folder, so when you run the second command, it just tries to find a `_package` folder inside, and then it will compress, and publish it.

## 2. Publishing the folder itself
`oc package your-component-folder`
`... store the _package folder somewhere to be published later`
`oc publish _package --skip-package`

In this case, the component is already gone, as we only stored the artifact of the package to be deployed/published at a later stage. So here the CLI will see that there's no _package folder inside of your path, so will try to find a package.json inside, and if there is one, check that we are on a compiled package (checking that theres an `oc.packaged = true` could do the trick here)

## 3. Publishing a compressed file
`oc package your-component-folder --compress`
`... store the tar.gz file somewhere to be published later`
`oc publish your-file.tar.gz`

Similar to the case 2 but here we also compress and we want to publish that file. In this case I don't see the point of skipping package, since CLI could already assume that if the path is of a compressed file, you just want to publish that.

Does this approach make sense?

PR #1188, opened 2021-06-14
add ability to override unpkg domain

### LINKED ISSUE #1187: Ability to override 'https://unpkg.com` domain at repository for templates.js
Consumers of OC, we would like to host unpkg.com react libraries on our domain so we minimize the number of domains, and control access.  unpkg.com is blocking Google Bots for some libraries and thus can effect SEO.

Browser client

## *When* does this occur?
When templates.js has a reference to unpkg.com

## *Where* on the platform does it happen?
Client, server-side rendered.

Publish a react component and set your user-agent to `Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)` you will see a 403.

Let the registry decide to allow to override the domain with a custom property for domain/path.

Not applicable

kmcrawford:
**Test plan (required)**

add the following you your registry configuration:
```
unpkgTemplateOverride: "https://mydomain.com/my-path",
```
Then publish a react component to your running registry, you will see the template.js file has been updated to replace: `https://unpkg.com` to the value in `conf.unpkgTemplateOverride`.  This function will only fire if there is a value in `unpkgTemplateOverride` and a `template.js` file exists.

I request some assistance on best way to write unit tests as this does update files.

**Closing issues**
closes #1187

### DIFF
--- a/src/registry/domain/repository.js
+++ b/src/registry/domain/repository.js
@@ -283,6 +283,29 @@ module.exports = function(conf) {
         );
       });
     },
+
+    overrideTemplateUnpkg: (pkgDetails, callback) => {
+      const templateJs = pkgDetails.outputFolder + '/template.js';
+      fs.access(templateJs, fs.F_OK, err => {
+        if (err) {
+          return callback(err);
+        }
+        fs.readFile(templateJs, 'utf8', (err, data) => {
+          if (err) {
+            return callback(err);
+          }
+          const result = data.replace(
+            /https:\/\/unpkg.com/g,
+            conf.unpkgTemplateOverride
+          );
+          fs.writeFile(templateJs, result, 'utf8', (err) => {
+            if (err) {
+              return callback(err);
+            }
+          });
+        });
+      });
+    },
     publishComponent: (
       pkgDetails,
       componentName,
@@ -358,6 +381,9 @@ module.exports = function(conf) {
             err => {
               if (err) {
                 return callback(err);
+              }
+              if (conf.unpkgTemplateOverride) {
+                repository.overrideTemplateUnpkg(pkgDetails, callback);
               }
               cdn.putDir(
                 pkgDetails.outputFolder,
--- a/src/registry/domain/repository.js
+++ b/src/registry/domain/repository.js
@@ -284,7 +284,7 @@ module.exports = function(conf) {
       });
     },
 
-    overrideTemplateUnpkg: (pkgDetails, callback) => {
+    overrideTemplates: (pkgDetails, templateOverrides, callback) => {
       const templateJs = pkgDetails.outputFolder + '/template.js';
       fs.access(templateJs, fs.F_OK, err => {
         if (err) {
@@ -294,11 +294,10 @@ module.exports = function(conf) {
           if (err) {
             return callback(err);
           }
-          const result = data.replace(
-            /https:\/\/unpkg.com/g,
-            conf.unpkgTemplateOverride
-          );
-          fs.writeFile(templateJs, result, 'utf8', (err) => {
+          templateOverrides.forEach(o => {
+            data = data.replace(o.findRegex, o.override);
+          });
+          fs.writeFile(templateJs, data, 'utf8', err => {
             if (err) {
               return callback(err);
             }
@@ -382,8 +381,12 @@ module.exports = function(conf) {
               if (err) {
                 return callback(err);
               }
-              if (conf.unpkgTemplateOverride) {
-                repository.overrideTemplateUnpkg(pkgDetails, callback);
+              if (conf.templateOverrides) {
+                repository.overrideTemplates(
+                  pkgDetails,
+                  conf.templateOverrides,
+                  callback
+                );
               }
               cdn.putDir(
                 pkgDetails.outputFolder,


matteofigus:
Thanks for the PR. 
I am trying to establish if this is the right place where to make this change. While this is a convenient place to automatically transform anything, I wonder if we should tackle this problem on the react template side instead? But I am not strongly opinionated.

As far as I can tell, the issue in the react template is around the [externals defined in the template](https://github.com/opencomponents/oc-template-react/blob/master/packages/oc-template-react/package.json#L39) that get bundled in by its compiler [as externals](https://github.com/opencomponents/oc-template-react/blob/master/packages/oc-template-react-compiler/lib/compileView.js#L46-L49) - maybe we could override those there? We could have a setting in the component itself to override and have the compiler to do something like:

```
const compile = (options, cb) => {
    const config = webpackConfigurator({
      viewPath: options.viewPath,
      externals: externals.reduce((externals, dep) => {
        if (componentPackage.oc.globalsBaseUrlOverride){
          externals[dep.name] = dep.global.replace(/https:\/\/unpkg.com/g, componentPackage.oc.globalsBaseUrlOverride);
        } else {
          externals[dep.name] = dep.global;
        }
        return externals;
      }, {}),
      publishFileName,
      production,
      buildIncludes: componentPackage.oc.files.template.buildIncludes || []
    });
    ...
```

kmcrawford:
@matteofigus I started down that path of overriding at compile time and having the client or component the ability to control the override.  The next thought is that this is could be more of a run-time variable that can be determined by registry.  In our environment we have about 30 OCs, maintained by multiple teams, and multiple registries.  Each registry could have a different serving domain and at that time was the thought to couple with the registry.

kmcrawford:
@matteofigus let me know how you would like to move forward with this as this is effecting our production environment

ricardo-devis-agullo:
I think the solution is very specific to one template (what if the template you are using points to a different cdn than unpkg?), and should be solved at the template level. Either that, or give an option to override externals completely at the package json. Something like 
```json
  "oc": {
    "files": {
      "template": {
        "src": "index.js",
        "type": "oc-template-react",
        "externals": [] // here you override
      },
    },
```

kmcrawford:
@ricardo-devis-agullo the only issue I have at build time is that I may want to have different runtime domains.  Thus I could keep all of the external packages on the local domain of the registry (on a CDN path).  I do like the thought of extending this for other domains, not only `unpkg`.  Let me look into a PR to see what that may look like.

kmcrawford:
@ricardo-devis-agullo , @matteofigus I have updated the oc to allow for many template overrides from the registry config: 
```
templateOverrides: [{
      findRegex: /https:\/\/unpkg.com/g,
      override: "https://mydomain/cdnpath"
    }]
```

ricardo-devis-agullo:
This feels a bit hacky, and doesn't cover scenarios where for example the domain where I host those statics doesn't follow the same path structure as the original url. I feel that a good solution would be to be able to have something like 
```json
What if the url that you want to override does'nt follow the same folder structure than the original one you want to replace with? I find that a simpler and more ergonomic approach is just to allow a list of externals to override that follow the same structure, so in your case, if you wanted to override the React and ReactDOM externals you could just have
```json
{
  "externals": [
    {
      "global": "React",
      "url": "yournewcdn.com/react.production.min.js"
    },
    {
      "global": "ReactDOM",
      "url": "yournewcdn.com/react.production.min.js"
    }
  ]
}
```

If that is not possible at the oc level, then I'd suggest to do it at the compiler level, or fork the compiler of your choice and change your urls.

kmcrawford:
@ricardo-devis-agullo using a regex allows to replace for different paths levels.  This is updated at deployment time to an oc registry vs package time.  This allows the domain of the externals to be variable for each registry.  For example if I deploy the same component to 3 different registries, I can serve the externals on the same domain that the registry is served off of, this allows the browser to not have to make an additional DNS lookup.   The immediate issue I am trying to solve is that unpkg.com is blocking Google bots, which is effecting our navigation to not load and thus hurting SEO.  If there is no `templateOverrides` in your registry configuration, the current functionality will remain unchanged.

pbazydlo:
Hey @kmcrawford I agree with @ricardo-devis-agullo that doing regex on the compiled template feels like a hack. It's quite confusing when you upload a certain code to registry but then it alters it.

Being explicit about this in `package.json` would make for a clearer interface. When deploying to different registries you could alter `package.json` right before `oc publish`.

kmcrawford:
I have attempted this route and it is not feasible.  If there is censuses that this feels "hacky", I need the ability to pass a custom function prior to uploading the template.js, therefore I will have the ability to solve my issue.

pbazydlo:
I'm probably missing something, sorry if I am :(
I'm looking at this bit of code in react template: https://github.com/opencomponents/oc-template-react/blob/45aae5cf5570df83c717b0d6e980ffc8e79fddd5/packages/oc-template-react-compiler/lib/compileView.js#L22

It seems to me that we have access to component `package.json` via `options.componentPackage`.
This file also seems to be where we're using externals:
https://github.com/opencomponents/oc-template-react/blob/45aae5cf5570df83c717b0d6e980ffc8e79fddd5/packages/oc-template-react-compiler/lib/compileView.js#L46-L48 and https://github.com/opencomponents/oc-template-react/blob/45aae5cf5570df83c717b0d6e980ffc8e79fddd5/packages/oc-template-react-compiler/lib/compileView.js#L90

We should be able to modify `compileView` file to check `componentPackage` if there are any overrides to apply to externals and create a good package from the get go. Do you think that could work?

As an aside it would be nice to be able to publish an already zipped package and there's some changes to publishing process that could be useful. Some of our piplines do `oc package`, unzip, modify the files, zip it up again and perform the same call on oc registry that `oc publish` would. It solved our immediate issues (and would probably help you as well), however that specifically is without question very hacky.

kmcrawford:
@pbazydlo that is interesting, so the thought would be that the CI adds the necessary overrides to the component's package.json before the `oc publish`, and the `compileView` would do the replacement.  That's something I could look into.

kmcrawford:
@pbazydlo I also +1 your `oc publish --no-package`

ricardo-devis-agullo:
There's an open issue about publishing without packaging here #769, which I've been working on lately, so perhaps will be available soon.

What we do right now on this is we package it, modify whatever we need and then we push directly to the registry by making a PUT call to the registry (bypassing the publish command). I haven't thought about the use case of wanting different domains for multiple registries, so I can see the value of that, but I think the solution should be addressed differently.

kmcrawford:
Closing in favor of https://github.com/opencomponents/oc-template-react/pull/647

PR #1191, opened 2021-08-14
Add cli argument to skip dependencies packaging

### LINKED ISSUE #769: Add cli argument to skip dependencies packaging

ricardo-devis-agullo:
Closes #769 

This makes it possible to directly publish an already packaged component. Of the 3 possible ways referenced on the issue I just did the first approach, which I guess is the more natural given the current flow of what publishing does, which is to to skip the packaging (creating the _package folder) and do the rest the same.

So
`oc publish --skipPackage your-component`
Will try to find a `_package` folder in `your-component` folder and publish it. It will fail with an error message if either the _package folder does not exist, or there isn't a package.json file inside (needed to get component info).

This helps for cases where you package/build it yourself to publish on a later stage (without needing to store all the dependencies needed for the bundling process).

### DIFF
--- a/src/resources/index.js
+++ b/src/resources/index.js
@@ -32,7 +32,7 @@ ${yellow(validMockObject)}`;
 
 const initSuccess = (componentName, componentPath) => {
   const success = `Success! Created ${componentName} at ${componentPath}`;
-  return `${green(success)} 
+  return `${green(success)}
 
 From here you can run several commands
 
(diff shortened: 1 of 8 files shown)


src/resources/index.js
```diff
@@ -173,6 +173,7 @@ module.exports = {
         'the version of used OC CLI is invalid. Try to upgrade OC CLI running {0}',
       PACKAGE_CREATION_FAIL: 'An error happened when creating the package: {0}',
       PACKAGING_FAIL: 'an error happened while packaging {0}: {1}',
+      PACKAGE_FOLDER_MISSING: 'Could not find a _package folder to publish',
```
pbazydlo, line 176: Would it be worth to start adding hints? ex. "make sure to run `oc package` first or drop the `--skip-package` parameter"

ricardo-devis-agullo: sure. done

pbazydlo (approved):

pbazydlo (approved):

Issue #512, opened 2017-06-06
registry metadata as API


mattiaerre:
# Description

it would be nice to have registry's metadata info exposed as an API; such as a list of available dependencies as well as plugins; @corlobepy @wmelani and myself will be more than happy to propose a PR that accommodates this need.

// cc @matteofigus @nickbalestra

matteofigus:
I think it's a good idea.
A couple of considerations about the implementation:

1) You want the api not to conflict with any component name. We use `~` with other endpoints as that's not allowed in a component name. So perhaps it could be something like `registry.com/~/api/plugins` or `registry.com/~api/plugins/` or maybe `registry.com/~registry/plugins`? (same would apply for dependencies).
2) Dependencies and plugins are private and we possibly don't want to expose them without authentication. I think the simpler thing to do now is to have those endpoints to return JSON in case of `discovery:true` setting, and a 401 (unauthorised) if `discovery:false`.
3) When designing the dependencies endpoint, we need to be prepared to the multi-versioned dependencies that we are gonna soon support
4) I have a separate issue #489 to show the available plugins via UI. One thing I was thinking about is that atm, the registry, apart from the name of the plugin and the actual handler, knows nothing more. I was thinking about making the interface of the plugins extensible to include some docs or links to them, so that exposing the list becomes a bit more useful from a developer's perspective. Perhaps we could open a separate story for that.

mattiaerre:
closing this issue 👋

PR #1194, opened 2021-08-21
Available plugins

### LINKED ISSUE #489: Expose plugins via the Registry UI
Possibly with some extra information, like some documentation and the API.

ricardo-devis-agullo:
Closes #489 

Adds a view to see the list of available plugins on the UI. Adds an extra optional parameter when registering to add a description to your plugin. This could be improved a lot by having real schematic instructions of how your function works (injecting/analyzing JSDoc on your plugin?) but that could be a separate story.

Also, as mentioned in issue #512, added also a JSON Api endpoint that gives the metadata list, but only if your service is set as discoverable.

### DIFF
--- a/src/registry/domain/plugins-initialiser.js
+++ b/src/registry/domain/plugins-initialiser.js
@@ -98,6 +98,8 @@ module.exports.init = function(pluginsToRegister, callback) {
     plugin.register.register(plugin.options || {}, dependencies, err => {
       const pluginCallback = plugin.callback || _.noop;
       pluginCallback(err);
+      // Overriding toString so implementation details of plugins do not
+      // leak to OC consumers
       plugin.register.execute.toString = () => plugin.description || '';
       registered[plugin.name] = plugin.register.execute;
       done(err);
--- a/src/registry/domain/plugins-initialiser.js
+++ b/src/registry/domain/plugins-initialiser.js
@@ -98,6 +98,7 @@ module.exports.init = function(pluginsToRegister, callback) {
     plugin.register.register(plugin.options || {}, dependencies, err => {
       const pluginCallback = plugin.callback || _.noop;
       pluginCallback(err);
+      plugin.register.execute.toString = () => plugin.description || '';
       registered[plugin.name] = plugin.register.execute;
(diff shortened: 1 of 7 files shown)


src/registry/routes/plugins.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+module.exports = function(conf) {
+  return function(req, res) {
+    if (conf.discovery) {
+      const plugins = Object.entries(conf.plugins).map(
+        ([pluginName, pluginFn]) => ({
+          name: pluginName,
+          description: pluginFn.toString()
```
pbazydlo, line 9: I'm not too fond of us abusing `toString` on plugin function for description. I'd rather we had a property `description` we could refer to

src/registry/domain/plugins-initialiser.js
```diff
@@ -98,6 +98,7 @@ module.exports.init = function(pluginsToRegister, callback) {
     plugin.register.register(plugin.options || {}, dependencies, err => {
       const pluginCallback = plugin.callback || _.noop;
       pluginCallback(err);
+      plugin.register.execute.toString = () => plugin.description || '';
       registered[plugin.name] = plugin.register.execute;
```
pbazydlo, line 102: I think it would be cleaner if we could do something like this:
```javascript
registered[plugin.name] = {
  execute:...
  description: ...
}
``` and then use those properties accordingly.

src/registry/routes/plugins.js
ricardo-devis-agullo: so abusing properties in functions it's fine then? :P

src/registry/domain/plugins-initialiser.js
ricardo-devis-agullo: but that's not what we pass to the registry

pbazydlo: we've clarified this on a call. we probably want to overload toString anyway. Comment is added to clarify this bit in the code as it is not obvious

pbazydlo (approved):

PR #1207, opened 2021-09-08
[BREAKING] Update oc-client-browser to 1.5.3

ricardo-devis-agullo:
This upgrade creates a small breaking change, since it updates jQuery from 1.11.2 to 3.6.0, and because `oc.$` exposes that jQuery, components that relied on the very old [deprecated methods](https://jquery.com/upgrade-guide/3.0/) from jQuery 1 won't work anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "multer": "^1.3.0",
     "nice-cache": "0.0.5",
     "oc-client": "3.2.12",
-    "oc-client-browser": "1.4.2",
+    "oc-client-browser": "1.5.3",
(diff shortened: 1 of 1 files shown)


pbazydlo:
How are we versioning this on oc? I guess 0.49? Do I remember correctly that oc version + changelog come as a separate PR?

ricardo-devis-agullo:
Yes, this will be 0.49. Oc version + changelog is done with a task on npm scripts, that among other things will also automatically push it to master, so you won't see a PR for that.

pbazydlo (approved):

PR #1209, opened 2021-09-14
remove wrap-cli-callback

ricardo-devis-agullo:
From what I can see `wrap-cli-callback` was only used for CLI commands where we didn't pass a callback. We can just pass it explicitly on `cli/index.js` when calling yargs, and then we don't need to wrap every command. This also makes it easier to be able to [universalify](https://github.com/RyanZim/universalify) those functions (it would get confused thinking it's a promise if sometimes they get called without a callback).


src/cli/facade/clean.js
```diff
@@ -3,7 +3,6 @@
 const read = require('read');
 
 const strings = require('../../resources/index');
-const wrapCliCallback = require('../wrap-cli-callback');
```
pbazydlo, line 6: by doing this aren't we breaking contract for things like this:
https://github.com/opencomponents/grunt-oc/blob/master/tasks/oc-mock.js

ricardo-devis-agullo: how? in that file it's explicitly passing a callback.

pbazydlo: we've chatted about this, the difference is about programmatic api, however we think that the use-case of not passing a callback in programmatic api does not make any sense. @matteofigus we're thinking of treating this as a non breaking change - is that ok?

matteofigus: yeah makes sense, thanks

PR #1210, opened 2021-09-15
remove new references on non-classes

ricardo-devis-agullo:
Removing `new` references on functions that return normal objects (not classes), since it's not needed.

### DIFF
--- a/src/cli/facade/dev.js
+++ b/src/cli/facade/dev.js
@@ -157,7 +157,7 @@ module.exports = function(dependencies) {
                 return callback(err);
               }
 
-              const registry = new oc.Registry({
+              const registry = oc.Registry({
                 baseUrl,
                 prefix: opts.prefix || '',
                 dependencies: dependencies.modules,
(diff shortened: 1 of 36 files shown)


src/cli/facade/dev.js
```diff
@@ -157,7 +157,7 @@ module.exports = function(dependencies) {
                 return callback(err);
               }
 
-              const registry = new oc.Registry({
+              const registry = oc.Registry({
```
pbazydlo, line 160: won't this change break how registry is initialised right now outside of this repository?

ricardo-devis-agullo: why would it? it's not a class, so what will be the difference outside?

pbazydlo: We've tested it together and it does not seem to affect the initalisation after all.

pbazydlo (approved):

matteofigus:
One thing about this, is making sure we update the docs like this https://github.com/opencomponents/oc/wiki#setup-a-registry

ricardo-devis-agullo:
Technically you can still call it with `new`, but yeah, I'll remove it too.

PR #1215, opened 2021-09-20
add default empty object for object keys

ricardo-devis-agullo:
_.keys() worked with `undefined` but `Object.keys` doesn't, so this adds defaults when needed


pbazydlo (approved):

pbazydlo:
When reviewing previous PR I've checked for differences, no-one mentioned that one :P

ricardo-devis-agullo:
👀

PR #1224, opened 2021-09-25
refactor more utils to typescript

ricardo-devis-agullo:
Refactor all utils except clean-require


src/utils/put.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  files.forEach(file => {
+    const fileName = path.basename(file);
+    form.append(fileName, fs.createReadStream(file));
+  });
+
+  options.headers = { ...headers, ...form.getHeaders() };
+
+  form.submit(options as any, (err, res) => {
```
pbazydlo, line 33: why do we need this any and on line 36?

ricardo-devis-agullo: clash betwen libraries, url.parse says protocol can be string | null and form.submit accepts string | undefined
its silly but im leaving anys that require refactors and I will tackle those on a second set of prs

pbazydlo: ok

pbazydlo (approved):

PR #1226, opened 2021-09-25
move registry middleware files to TS

ricardo-devis-agullo:


src/registry/middleware/index.ts
```diff
@@ -41,7 +46,7 @@ module.exports.bind = function(app, options) {
   }
 
   if (options.local) {
-    app.use(errorhandler({ dumpExceptions: true, showStack: true }));
+    app.use(errorhandler());
```
pbazydlo, line 49: does `errorhandler()` have those previous settings as defaults?

ricardo-devis-agullo: typescript complained those types don't exist. I checked the source code, those options dont exist

pbazydlo (approved):

PR #1228, opened 2021-09-25
move routes helpers to TS

ricardo-devis-agullo:


src/registry/routes/helpers/apply-default-values.ts
```diff
@@ -0,0 +1,21 @@
...
+  expectedParameters: Dictionary<OcParameter> = {}
+): Dictionary<string | number | boolean> {
+  const optionalParametersWithDefaults = Object.entries(
+    expectedParameters
+  ).filter(([, p]) => !p.mandatory && typeof p.default !== 'undefined');
+
+  optionalParametersWithDefaults.forEach(
+    ([expectedParameterName, expectedParameter]) => {
+      const param = requestParameters[expectedParameterName];
+      if (param == null) {
```
pbazydlo, line 14: previous version of code let `undefined` through at the filter stage and for `undefined` paramters it provided defaults. I think this will not provide default is parameter is `undefined`

src/registry/routes/helpers/get-components-history.ts
```diff
@@ -0,0 +1,32 @@
...
+interface ComponentHistory {
+  name: string;
+  publishDate: number;
+  version: string;
+}
+
+export default function getComponentsHistory(history: ComponentsDetails) {
+  const result: ComponentHistory[] = [];
+
+  for (const [name, versions] of Object.entries(history.components)) {
```
pbazydlo, line 13: why is the order in array switched when compared with original?

ricardo-devis-agullo: Because lodash .each/.forEach puts the value first and then the key
https://lodash.com/docs/4.17.15#forEach
But thats not the case with Object.entries

src/registry/routes/helpers/apply-default-values.ts
ricardo-devis-agullo: `param == null` checks for both null and undefined
one of the few cases where the double equality operator is useful and predictable

pbazydlo: oh it's the double equality... I'm wondering if it's just me, but it be easier for me to understand this if we used `===` and had two conditions like before. double equality operator breaks my mind :P it could be a small follow up PR if it's hard for you to change this branch

ricardo-devis-agullo: done

pbazydlo (approved):

PR #1273, opened 2021-10-31
move nested renderer to promises

ricardo-devis-agullo:
Transforming nested-renderer to promises. When exported in get-component they will be universalified, because they are exported on the ocs, so to not break we need to keep supporting callback cased calls

### DIFF
--- a/src/registry/domain/nested-renderer.ts
+++ b/src/registry/domain/nested-renderer.ts
@@ -1,145 +1,87 @@
-import async from 'async';
-import _ from 'lodash';
-
 import settings from '../../resources/settings';
 import strings from '../../resources';
 import { Config } from '../../types';
+import {
+  GetComponentResult,
+  RendererOptions
+} from '../routes/helpers/get-component';
 
-type Cb = Callback<string, string>;
-type Options = {
+interface Options {
+  ip?: string;
   version?: string;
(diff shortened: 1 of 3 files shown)


src/registry/domain/nested-renderer.ts
```diff
@@ -1,145 +1,87 @@
...
-        strings.errors.registry.NESTED_RENDERER_COMPONENTS_IS_NOT_VALID
-      );
-    }
+}
 
-    validate.callback(params.callback);
-  }
-};
+export default function nestedRenderer(
+  rendererCb: (
```
pbazydlo, line 18: How about `rendererWithCallback`?

test/unit/registry-domain-nested-renderer.js
```diff
@@ -265,22 +211,27 @@ describe('registry : routes : helpers : nested-renderer', () => {
...
+          nestedRenderer
+            .renderComponents(
+              [
+                { name: '404-component' },
+                { name: 'another-not-existent-component' }
+              ],
+              {}
+            )
+            .then(res => {
+              console.log('RES:', res);
```
pbazydlo, line 223: let's clean up those

pbazydlo (approved):

PR #1276, opened 2021-11-12
[CLI-FEATURE] filter components to run when running oc dev

ricardo-devis-agullo:
When running `oc dev` on a folder, it will try to register and package all the oc components that it can find, but that sometimes can be undesirable on the following circumstances:

1) You don't want to have to package multiple components that you are not testing at that moment (more time compiling).
2) You want some of those components to be retreived from a fallback registry.

For that, I'm adding a new option `--components` in `oc dev` that takes a list of folders to run against. So given that you are on a folder with

`component1, component2, component3, component4`

When running the command

`oc dev . --components component1 component3`

It will only package and show on the local registry the components 1 and 3.

### DIFF
--- a/src/cli/commands.ts
+++ b/src/cli/commands.ts
@@ -52,6 +52,10 @@ export default {
           boolean: false,
           description: 'Url prefix for registry server',
           default: ''
+        },
+        components: {
+          array: true,
+          description: 'Filter only the components that you want to be included'
         }
       },
       usage: 'Usage: $0 dev <dirPath> [port] [baseUrl] [options]'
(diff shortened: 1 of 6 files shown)


src/cli/commands.ts
```diff
@@ -52,6 +52,10 @@ export default {
           boolean: false,
           description: 'Url prefix for registry server',
           default: ''
+        },
+        components: {
+          array: true,
+          description: 'Filter only the components that you want to be included'
```
pbazydlo, line 58: How about: `List of component names that you want to be packaged and exposed on dev registry`

ricardo-devis-agullo: done

pbazydlo (approved):

PR #1277, opened 2021-11-13
[CLI-FEATURE] add registries options for oc publish

ricardo-devis-agullo:
Right now, to publish to registries you need to have set a `oc.json` file with your registries. This works well when you only publish to one registry, or you always publish to all the registries at the same time (so you can have your oc.json versioned in your repo and just use it).

The problem is that that solution doesn't work as well when your deployment setup is to publish to a different registry depending on your environment (1 or more registries on each environment). In that case, for every deployment, you have to go and edit the oc.json before being able to call `oc publish`.

For those cases, I'm adding a new option `--registries` that allows you to pass a list of registries that will override your oc.json configuration. So when you want to publish you can just do

```
oc publish your-component --registries http://www.myregistry.com https://www.myregistry2.com
```
without needing a oc.json

### DIFF
--- a/src/cli/commands.ts
+++ b/src/cli/commands.ts
@@ -132,7 +132,7 @@ export default {
         registries: {
           array: true,
           description:
-            'registries to publish, overriding your configuration in oc.json'
+            'List of registries to publish to. This setting will take precedence over oc.json file'
         }
       },
       description: 'Publish a component',
--- a/src/cli/commands.ts
+++ b/src/cli/commands.ts
@@ -128,6 +128,11 @@ export default {
           boolean: true,
           description: 'Skip packaging step',
           default: false
+        },
+        registries: {
+          array: true,
+          description:
+            'registries to publish, overriding your configuration in oc.json'
         }
       },
       description: 'Publish a component',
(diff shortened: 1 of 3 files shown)


src/cli/commands.ts
```diff
@@ -128,6 +128,11 @@ export default {
           boolean: true,
           description: 'Skip packaging step',
           default: false
+        },
+        registries: {
+          array: true,
+          description:
+            'registries to publish, overriding your configuration in oc.json'
```
pbazydlo, line 135: `List of registries to publish to. This setting will take precedence over oc.json file.`

ricardo-devis-agullo: done

test/unit/cli-facade-publish.js
```diff
@@ -101,6 +102,20 @@ describe('cli : facade : publish', () => {
         });
       });
 
+      it('should take precedence the registries set through options', done => {
```
pbazydlo, line 105: `should take precedence over the...`

ricardo-devis-agullo: done

PR #1283, opened 2021-12-15
add "open" link to component preview page

qbikez:
Add a "Open" link next to "Preview" to open component preview in a new tab. 

The component preview frame is small and not very usable for testing. I always find myself copying the link, adding "~preview" and pasting it to a new tab. Now the link can do it for me.

image

### DIFF
--- a/src/registry/views/static/info.ts
+++ b/src/registry/views/static/info.ts
@@ -7,7 +7,7 @@ oc.cmd.push(function() {
     window.location = thisComponentHref + $(this).val() + '/~info';
   });
 
-  $('.refresh-preview').click(function() {
+  refreshPreview = function() {
     var splitted = $('#href')
         .val()
         .split('?'),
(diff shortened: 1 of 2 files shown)


src/registry/views/static/info.ts
```diff
@@ -7,7 +7,7 @@ oc.cmd.push(function() {
     window.location = thisComponentHref + $(this).val() + '/~info';
   });
 
-  $('.refresh-preview').click(function() {
+  refreshPreview = function() {
```
ricardo-devis-agullo, line 10: I think you want to do either `var refreshPreview  = function() {` or `function refreshPreview() {` so it does not become a global variable.

qbikez: ok, fixed that to `var refreshPreview = ...`

ricardo-devis-agullo (approved):

PR #1287, opened 2022-01-17
[INTERNAL] reduce lodash usage on test/code

ricardo-devis-agullo:


test/unit/cli-domain-handle-dependencies-get-compiler.js
```diff
@@ -33,7 +32,7 @@ describe('cli : domain : handle-dependencies : get-compiler', () => {
       }
     ).default;
 
-    getCompiler(_.cloneDeep(options))
+    getCompiler(options)
```
pbazydlo, line 35: why are we not cloning anymore?

ricardo-devis-agullo: because options are created per test on this same function and only used to being passed to this function, so it was absolutely useless cloning

pbazydlo: makes sense

pbazydlo (approved):

PR #1315, opened 2022-12-08
include the entire console api in the production stub

### LINKED ISSUE #1316: Console is not fully defined in production
## _Who_ is the bug affecting?

Component authors/creators.

## _What_ is affected by this bug?

The ability to server-side render components.

## _When_ does this occur?

When trying to server render a component in production.

## _Where_ on the platform does it happen?

In the registry in production only. This issue is not present locally as the "real" console object is passed into the `vm` context. This makes this problem particularly dangerous.

## _How_ do we replicate the issue?

1. Add `console.warn` anywhere inside `server.js`. 
2. Build component and run inside production (non-local) registry.
3. Observe `console.warn is not a function` error when trying to fetch/get component from registry.

`console` api should be fully stubbed inside component server execution. I've drafted a PR: https://github.com/opencomponents/oc/pull/1315 that I hope details what I think should happen.

oc@0.49.6
node@16.18.1

mungodewar:
Usage of `console.warn` will prevent the entire component from rendering in the production oc-registry. This PR addresses this by fully specifying a `console` stub in production.

**Closing issues**
closes #1316

### DIFF
--- a/src/registry/routes/helpers/get-component.ts
+++ b/src/registry/routes/helpers/get-component.ts
@@ -491,7 +491,27 @@ export default function getComponent(conf: Config, repository: Repository) {
                       (...args: unknown[]) => unknown
                     >
                   },
-                  console: conf.local ? console : { log: _.noop },
+                  console: conf.local ? console : {
+                    assert: _.noop,
+                    clear: _.noop,
+                    count: _.noop,
+                    countReset: _.noop,
+                    debug: _.noop,
+                    dir: _.noop,
+                    dirxml: _.noop,
+                    error: _.noop,
+                    group: _.noop,
+                    groupCollapsed: _.noop,
+                    groupEnd: _.noop,
+                    info: _.noop,
+                    log: _.noop,
+                    table: _.noop,
+                    time: _.noop,
+                    timeEnd: _.noop,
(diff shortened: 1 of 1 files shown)


src/registry/routes/helpers/get-component.ts
```diff
@@ -491,7 +491,27 @@ export default function getComponent(conf: Config, repository: Repository) {
...
+                    groupEnd: _.noop,
+                    info: _.noop,
+                    log: _.noop,
+                    table: _.noop,
+                    time: _.noop,
+                    timeEnd: _.noop,
+                    timeLog: _.noop,
+                    trace: _.noop,
+                    warn: _.noop,
+                    },
```
ricardo-devis-agullo, line 514: I think you could replace this by just
```typescript
console: conf.local ? console : Object.fromEntries(Object.keys(console).map(key => [key, _.noop])),
```

mungodewar: Happy to use that solution!

mungodewar: ⚡ Fixed up

ricardo-devis-agullo (approved):

PR #1321, opened 2023-01-12
Put cdn.putDir logic inside OC

ricardo-devis-agullo:
This puts the `putDir` logic back on OC. The reason is that
1) The logic is the same for all storage adapters (since no storage has really the concept of directories, it just does a loop of putFiles)
2) There's specific OC logic (making server.js and .env private) that doesn't belong on storage adapters, because they should be agnostic about OC and just have the simple logic of managing files on a storage.


ricardo-devis-agullo:
Closing stale PR from 2023. putDir/storage direction has moved on; not worth rebasing this giant churn.

PR #1327, opened 2023-03-08
showing all error when publishing fails

### LINKED ISSUE #1326: OC doesn't show proper message when publishing fails
## _Who_ is the bug affecting?

- component creators

## _What_ is affected by this bug?

- OC publishing client api

## _When_ does this occur?

- Most of the times when components fail during publishing.

## _Where_ on the platform does it happen?

- Linux

## _How_ do we replicate the issue?

- Using [OC S3 issue](https://github.com/opencomponents/oc/issues/1325).

- If oc publishing fails, it should show us proper error message saying why it has failed instead of empty error message
```
An error happened when publishing the component: {}
```

- OC@0.49.23, Node 16.X.X, Linux

## Other Comments
- Proposed solution is to stringify the [error object](https://github.com/opencomponents/oc/blob/b03bca52dc758c2bbd31aa80cb044677c613b942/src/registry/routes/publish.ts#L69) instead of extracting message field which can change the name (err.msg or err.message).
- Sample patch file
```
--- ./oc-patch/oc-registry-routes/publish_original.js	2023-03-01 11:37:29.000000000 +0000
+++ ./oc-patch/oc-registry-routes/publish_modified.js	2023-03-01 11:40:50.000000000 +0000
@@ -70,25 +70,35 @@
                 res.status(200).json({ ok: true });
             }
             catch (err) {
+                console.error('Patched OC->Registry->Routes->Publish: ', JSON.stringify(err));
+                let errorMessage;
+                if(err.msg) {
+                    errorMessage = err.msg;
+                } else if (err.message){
+                    errorMessage = err.message;
+                } else {
+                    errorMessage = JSON.stringify(err);
+                }
+
                 if (err.code === 'not_allowed') {
-                    res.errorDetails = `Publish not allowed: ${err.msg}`;
-                    res.status(403).json({ error: err.msg });
+                    res.errorDetails = `Publish not allowed: ${errorMessage}`;
+                    res.status(403).json({ error: err });
                 }
                 else if (err.code === 'already_exists') {
-                    res.errorDetails = `Component already exists: ${err.msg}`;
-                    res.status(403).json({ error: err.msg });
+                    res.errorDetails = `Component already exists: ${errorMessage}`;
+                    res.status(403).json({ error: err });
                 }
                 else if (err.code === 'name_not_valid') {
-                    res.errorDetails = `Component name not valid: ${err.msg}`;
-                    res.status(409).json({ error: err.msg });
+                    res.errorDetails = `Component name not valid: ${errorMessage}`;
+                    res.status(409).json({ error: err });
                 }
                 else if (err.code === 'version_not_valid') {
-                    res.errorDetails = `Component version not valid: ${err.msg}`;
-                    res.status(409).json({ error: err.msg });
+                    res.errorDetails = `Component version not valid: ${errorMessage}`;
+                    res.status(409).json({ error: err });
                 }
                 else {
-                    res.errorDetails = `Publish failed: ${err.msg}`;
-                    res.status(500).json({ error: err.msg });
+                    res.errorDetails = `Publish failed: ${errorMessage}`;
+                    res.status(500).json({ error: err });
                 }
             }
         }


```

ashmeet-kandhari:
Relates to #1326 

- Putting the whole error object in error response when publishing fails.
This is to avoid empty error responses as mentioned in issue.

`closes #1326 `.

### DIFF
--- a/src/registry/routes/publish.ts
+++ b/src/registry/routes/publish.ts
@@ -64,20 +64,28 @@ export default function publish(repository: Repository) {
         );
         res.status(200).json({ ok: true });
       } catch (err: any) {
+        let errorMessage;
+        if(err.msg) {
+          errorMessage = err.msg;
+        } else if (err.message){
+          errorMessage = err.message;
+        } else {
+          errorMessage = JSON.stringify(err);
+        }
         if (err.code === 'not_allowed') {
-          res.errorDetails = `Publish not allowed: ${err}`;
+          res.errorDetails = `Publish not allowed: ${errorMessage}`;
           res.status(403).json({ error: err });
         } else if (err.code === 'already_exists') {
-          res.errorDetails = `Component already exists: ${err}`;
+          res.errorDetails = `Component already exists: ${errorMessage}`;
           res.status(403).json({ error: err });
         } else if (err.code === 'name_not_valid') {
-          res.errorDetails = `Component name not valid: ${err}`;
+          res.errorDetails = `Component name not valid: ${errorMessage}`;
           res.status(409).json({ error: err });
         } else if (err.code === 'version_not_valid') {
-          res.errorDetails = `Component version not valid: ${err}`;
+          res.errorDetails = `Component version not valid: ${errorMessage}`;
           res.status(409).json({ error: err });
         } else {
-          res.errorDetails = `Publish failed: ${err}`;
+          res.errorDetails = `Publish failed: ${errorMessage}`;
           res.status(500).json({ error: err });
         }
       }
--- a/src/registry/routes/publish.ts
+++ b/src/registry/routes/publish.ts
@@ -65,20 +65,20 @@ export default function publish(repository: Repository) {
         res.status(200).json({ ok: true });
       } catch (err: any) {
         if (err.code === 'not_allowed') {
-          res.errorDetails = `Publish not allowed: ${err.msg}`;
-          res.status(403).json({ error: err.msg });
+          res.errorDetails = `Publish not allowed: ${err}`;
+          res.status(403).json({ error: err });
         } else if (err.code === 'already_exists') {
-          res.errorDetails = `Component already exists: ${err.msg}`;
-          res.status(403).json({ error: err.msg });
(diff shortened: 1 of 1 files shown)


src/registry/routes/publish.ts
```diff
@@ -65,20 +65,20 @@ export default function publish(repository: Repository) {
         res.status(200).json({ ok: true });
       } catch (err: any) {
         if (err.code === 'not_allowed') {
-          res.errorDetails = `Publish not allowed: ${err.msg}`;
-          res.status(403).json({ error: err.msg });
+          res.errorDetails = `Publish not allowed: ${err}`;
```
ricardo-devis-agullo, line 68: I think the proposed solutions included more lines to transform the errormessage

ashmeet-kandhari: Hey Ricardo,

I should have mentioned it while raising the PR.
Anyways in the proposed solution
I was checking the `error object` if it has `msg` field or `message` field, if none then I was `JSON.stringify`ing the object and setting it everywhere

But it made sense to me, to just send the whole error object.

Let me know if you think otherwise or why should we not do this.

ricardo-devis-agullo: Well I'm thinking that we probably shouldn't return a stack trace on a production environment API

ashmeet-kandhari: Makes sense.
I have made the changes that match the same as I had proposed in the issue.

ashmeet-kandhari: Do you think we should also include logging errors ?

ashmeet-kandhari: Do you think, will it be better to use `res.conf.debug` instead of local?

ricardo-devis-agullo: There is no conf.debug property. Local should be fine.

ashmeet-kandhari: Cool, I have made the changes

ricardo-devis-agullo (approved):

ricardo-devis-agullo:
LGTM

Issue #1338, opened 2023-06-04
Add --dry-run option to publish


ricardo-devis-agullo:
The idea is [the same as in npm](https://docs.npmjs.com/cli/v9/commands/npm-publish#dry-run). A dry-run publish is useful for having it on pipelines for OCs so you know your changes are going to pass validation.

PR #1339, opened 2023-06-04
add dryrun on CLI and Registry

ricardo-devis-agullo:
Closes issue #1338

### DIFF
--- a/src/cli/facade/publish.ts
+++ b/src/cli/facade/publish.ts
@@ -27,12 +27,14 @@ const publish = ({
     async (opts: {
       componentPath: string;
       skipPackage?: boolean;
+      dryRun?: boolean;
       username?: string;
       password?: string;
       registries?: string[];
     }): Promise<void> => {
       const componentPath = opts.componentPath;
       const skipPackage = opts.skipPackage;
+      const dryRun = opts.dryRun;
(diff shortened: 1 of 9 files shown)


src/cli/facade/publish.ts
```diff
@@ -152,7 +154,10 @@ const publish = ({
       ) => {
         for (const registryUrl of registryLocations) {
           const registryNormalised = registryUrl.replace(/\/$/, '');
-          const componentRoute = `${registryNormalised}/${component.name}/${component.version}`;
+          let componentRoute = `${registryNormalised}/${component.name}/${component.version}`;
+          if (dryRun) {
+            componentRoute += '?checkOnly';
```
pbazydlo, line 159: should we call it `dryRun`, or `dry` as well?

ricardo-devis-agullo: The thing is to me, dryRun is context dependant. When you are on the CLI you just dry running, but when you call an API there is not really a concept of a dry run, and something like "only check" makes more sense, thats why the two terms for the two contexts: CLI / API.

pbazydlo: hmm... onlyCheck feels quite clunky to be honest, where, in fact, we're asking the server to do a dry-run as well, setting expectation of no side-effects

PR #1373, opened 2024-08-05
clean up corrupted versions after 1 minute

ricardo-devis-agullo:
This makes usage of the new `removeDir` method available on the storage adapters to clean up corrupted components (that is, components that failed to be published while in the middle of it, so only some files got to storage).

It will wait 1 minute before cleaning, and after a minute it will check again and if its still incomplete, it will delete it. This is in case a new instance of a registry boots up and thinks a component that its still in the middle of publishing to be corrupted.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
         "dotenv": "^16.4.5",
         "errorhandler": "^1.5.1",
         "express": "^4.19.2",
+        "fastq": "^1.17.1",
         "form-data": "^4.0.0",
         "fs-extra": "^11.2.0",
         "getport": "^0.1.0",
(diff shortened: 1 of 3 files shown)


ricardo-devis-agullo:
Closing as stale / superseded.

This PR auto-GCs half-published (corrupted) version folders from storage after a 1-minute re-check. That problem is real on the classic directory-scan components list, but the direction of travel is the metadata store adapter, which is a better place to own component version integrity and cleanup than bolting delayed `removeDir` onto the legacy components-list scan.

Not pursuing more work on this path — if cleanup is still needed later, it should live with the metadata store rather than this approach.

PR #1420, opened 2025-07-29
Add option to rate limit the publish endpoint

ricardo-devis-agullo:
This adds an option to rate-limit the publish endpoint. By default it will use a memory one with lru-cache, but it's extendable by passing an object that follows the interface described so you could plug-in Redis or whatever library you want. I'm leaning towards enabling this by default

### DIFF
--- a/src/registry/middleware/publish-rate-limit.ts
+++ b/src/registry/middleware/publish-rate-limit.ts
@@ -1,0 +1,67 @@
+import type { NextFunction, Request, Response } from 'express';
+import type { Config } from '../../types';
+import MemoryRateLimitStore from '../domain/memory-rate-limit-store';
+
+const DEFAULT_WINDOW_MS = 15 * 60 * 1000; // 15 minutes
+const DEFAULT_MAX_HITS = 100;
+const DEFAULT_MAX_CACHE_SIZE = 1000; // Maximum number of rate limit entries to keep in memory
+
+function defaultKeyGenerator(req: Request): string {
+  return `${req.ip}:${req.user ?? 'anon'}`;
+}
+
+export default function createPublishRateLimiter(conf: Config) {
+  const rateLimitConfig = conf.publishRateLimit ?? {};
+  const windowMs = rateLimitConfig.windowMs ?? DEFAULT_WINDOW_MS;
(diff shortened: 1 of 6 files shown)


src/registry/middleware/publish-rate-limit.ts
```diff
@@ -0,0 +1,68 @@
...
+const DEFAULT_WINDOW_MS = 15 * 60 * 1000; // 15 minutes
+const DEFAULT_MAX_HITS = 100;
+const DEFAULT_MAX_CACHE_SIZE = 1000; // Maximum number of rate limit entries to keep in memory
+
+function defaultKeyGenerator(req: Request): string {
+  return `${req.ip}:${req.user ?? 'anon'}`;
+}
+
+export default function createPublishRateLimiter(conf: Config) {
+  const rateLimitConfig = conf.publishRateLimit ?? {};
```
pbazydlo, line 14: rather than gathering things like windowMs, maxHits, maxCacheSize, etc. on global level wouldn't it be better to leave it purely to the rate limit store?

So you could provide in config a factory function that would provide those values to the store directly. For default store you would provide those values on line 22, but not configurable.

src/registry/router.ts
```diff
@@ -28,6 +29,8 @@ export function create(app: Express, conf: Config, repository: Repository) {
   };
 
   const prefix = conf.prefix;
+  const publishRateLimiter = createPublishRateLimiter(conf);
+  console.log('publishRateLimiter', publishRateLimiter);
```
pbazydlo, line 33: I guess this console.log is not intended to be here

src/registry/router.ts
```diff
@@ -50,6 +53,7 @@ export function create(app: Express, conf: Config, repository: Repository) {
   } else {
     app.put(
       `${prefix}:componentName/:componentVersion`,
+      // publishRateLimiter,
```
pbazydlo, line 56: is this meant to be commented out?

ricardo-devis-agullo: No I forgot while testing

src/registry/router.ts
ricardo-devis-agullo: Yup will delete

PR #1480, opened 2026-05-27
perf(get-component): cache acceptLanguageParser.parse results

ricardo-devis-agullo:
acceptLanguageParser.parse is called on every data-provider request with
the same Accept-Language header value. Since it's a pure function (same
input → same output), cache the result using the existing nice-cache instance
keyed by the raw header string.

This avoids repeated parsing of identical Accept-Language headers across
concurrent and sequential requests from the same browser.

### DIFF
--- a/packages/oc/src/registry/routes/helpers/get-component.ts
+++ b/packages/oc/src/registry/routes/helpers/get-component.ts
@@ -141,6 +141,14 @@ export default function getComponent(conf: Config, repository: Repository) {
   });
   const convertPlugins = pluginConverter(conf.plugins);
 
+  const parseAcceptLanguage = (header: string) => {
+    const cached = cache.get('accept-language', header);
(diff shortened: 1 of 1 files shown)


ricardo-devis-agullo:
Closing as superseded.

The original goal (avoid re-parsing the same Accept-Language header on every data-provider call) is already covered better on master by the lazy per-request getter introduced in the Jul 23 perf work (`get acceptLanguage` + `parsedAcceptLanguage ??=` in `get-component.ts`):

- parse only if the data provider actually reads `context.acceptLanguage`
- at most one parse per request

Cross-request caching of parsed headers (this PR’s approach via nice-cache) is a smaller incremental win and not worth rebasing/conflict-resolving on top of that. If we ever want process-wide caching later, it can be a fresh bounded-cache change on the getter — not this branch.

PR #1483, opened 2026-06-13
perf(registry): bound component file cache with lru-cache and add cacheMaxSize config

ricardo-devis-agullo:
Replace the unbounded nice-cache in get-component with lru-cache so
compiled templates, data providers and env files are evicted after
reaching cacheMaxSize (default 100). Also avoid shallow-cloning every
component detail map in components-details when nothing has changed.

- Add lru-cache@^10.4.3 dependency, remove nice-cache
- Add cacheMaxSize config option with sanitiser default
- Update tests for the new config default


ricardo-devis-agullo:
@copilot resolve the merge conflicts in this pull request

PR #1494, opened 2026-06-13
fix: registry security and traversal vulnerabilities

MatrixNeoKozak:
1. Summary of changes: Added mandatory component name validation to all registry repository methods to prevent directory traversal attacks and refactored error handling in the `getComponent` method to avoid exposing detailed internal error information.

2. Rationale behind the implementation: The registry repository lacked validation for the `componentName` parameter in several functions (`getComponent`, `getComponentInfo`, `getDataProvider`, `getEnv`), which could allow for directory traversal attacks if a malicious component name was provided. Additionally, the original error handling in `getComponent` leaked internal error details via `errorToString(err)`, violating the security policy to only return generic error messages to API consumers.

3. Technical impact analysis: Improves the security posture of the registry by preventing unauthorized access to files outside the designated component directory and reducing the amount of sensitive internal information exposed in error scenarios.

### DIFF
--- a/packages/oc/src/registry/domain/repository.ts
+++ b/packages/oc/src/registry/domain/repository.ts
@@ -154,6 +154,12 @@ export default function repository(conf: Config) {
       componentName: string,
       componentVersion?: string
     ): Promise<Component> {
+      if (!validator.validateComponentName(componentName)) {
+        throw {
+          code: strings.errors.registry.COMPONENT_NAME_NOT_VALID_CODE,
+          msg: strings.errors.registry.COMPONENT_NAME_NOT_VALID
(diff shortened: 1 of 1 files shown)


ricardo-devis-agullo:
I’m not convinced this PR addresses a real exploitable vulnerability in the registry. A registry always uses a cloud adapter with an object storage, where ../ is treated as part of the object key, not as filesystem traversal. Not to mention that all files on that container are public (and you can't escape the container) The only place this  resembles directory traversal is local dev mode, which is not exposed to untrusted users. Can you provide a concrete exploit example that would work against any of the storage adapters?

PR #1496, opened 2026-06-14
Add Azurite-backed memory stress benchmarks to OC

ricardo-devis-agullo:
Introduces Azurite-backed benchmark scenarios that simulate a large component registry, generate heavy component metadata, and track peak heap/RSS usage to validate memory consumption improvements.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1310,6 +1310,551 @@
         "node": ">= 12.0.0"
       }
     },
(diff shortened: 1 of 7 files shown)


ricardo-devis-agullo:
Closing — large Azurite benchmark harness is not a current priority. Can revisit as a fresh PR if we need memory stress benchmarks later.

PR #1511, opened 2026-07-07
Cache neutral facade once per request in Express adapter

ricardo-devis-agullo:
## Summary
Stacked on #1510 (phase 4). Tackles the per-call request/response facade wrapping overhead introduced in the phase-4 Express adapter.

- Cache `OcResponse` on the underlying res via a `Symbol`; the handler chain (~6-7 wraps/request) now reuses one instance instead of re-running `Object.assign` per call.
- Hoist the `res.stream` closure to a single shared function (`readable.pipe(this.raw)`) so a fresh closure is no longer allocated per wrapped handler.
- Memoize normalized `req.params` keyed on the identity of `req.params`; recompute only when Express reassigns it (new router layer), collapse the per-handler `{ ...params }` spread + splat-join to once per route layer. Initialize `raw`/`routeId` once.
- Synchronous `next()` fast path in `toExpressHandler`: only defer via a microtask when the handler actually returns a thenable. Sync middlewares (`res.conf`, `cors`, `baseUrl`, `discovery`) and sync route handlers no longer pay the unconditional `Promise.resolve(...).then(next)` microtask.
- Add acceptance coverage for the local-mode static route (\`GET /:componentName/:componentVersion/static/*splat\`) — the gap that let the phase-4 splat crash slip through.

Behaviour preserved: `use()` handlers continue the chain if they didn't send; route handlers never auto-`next()`; `req.params.splat` stays a slash-joined string. Public contract (`Registry().app`, `{ app, server }`, Express augmentation) untouched. Establishes the convert-once + cache pattern the future Fastify adapter (phase 6) should mirror via `decorateRequest`/`decorateReply` + `onRequest`.

#### Test plan
- [x] \`npx @biomejs/biome check --write \"packages/oc/src/registry\"\`
- [x] \`npm run build -w oc\`
- [x] \`npm run test -w oc\` — 930 passing (927 baseline + 3 new)

### DIFF
--- a/packages/oc/src/registry/domain/http-server/express-adapter.ts
+++ b/packages/oc/src/registry/domain/http-server/express-adapter.ts
@@ -16,10 +16,29 @@ import type {
 } from './types';
 
 const expressMiddleware = Symbol('expressMiddleware');
+const ocResponseSym = Symbol('ocResponse');
+const ocParamsSym = Symbol('ocParams');
 
 type WrappedOcHandler = OcHandler & {
   [expressMiddleware]?: ExpressMiddleware;
 };
+
+interface ParamsMemo {
+  source: object;
+}
+
+function stream(this: OcResponse, readable: NodeJS.ReadableStream): void {
+  readable.pipe(this.raw);
+}
+
+function normaliseParams(raw: Record<string, unknown>): Record<string, string> {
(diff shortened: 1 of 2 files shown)


ricardo-devis-agullo:
@copilot why are the pipeline ci builds not running on this PR?

PR #1513, opened 2026-07-07
Address phase 6 review: fix packaging, types, and Fastify behavior

ricardo-devis-agullo:
**Adapter implementation:**
- Implement HttpServerAdapter interface against Fastify
- Body parsing with qs urlencoded semantics
- Cookie support
- Multipart uploads normalized to OC upload file shape
- Request timing, logging, local error handling
- Strong ETag/304 support
- HEAD/OPTIONS parity with Express
- Re-export neutral HTTP adapter types from oc package root

**Package & dependencies:**
- Fix package entry points (tsconfig rootDir, exclude dist/test from published files)
- Move oc to peerDependency (>=0.50.56) and devDependency to avoid pulling Express into adapter installs

**Neutral types:**
- Replace Express/Multer type aliases with structural CookieOptions/UploadedFile to avoid external type dependency leakage

**Behavior fixes:**
- Cookies: translate Express maxAge (ms→s) and default Path=/; fail on signed cookies
- Listen: default host to 0.0.0.0, accept host option, report invalid ports via callback
- Body: add gzip/deflate/brotli inflation via preParsing with receivedEncodedLength
- Logging: guard skip function against missing res.conf in adapter and core middleware
- Multipart: throw LIMIT_FIELD_VALUE-style 413 on field truncation
- OPTIONS: scope Allow header to matching routes when possible
- Fallthrough: 404 only on empty handler arrays
- Response: track hasSent immediately to avoid false fallthroughs; remove default end() content-type

**Tests:**
- Adapter tests: ETag/304, OPTIONS, urlencoded qs, cookies, wildcard params, connect middleware, streaming, multipart, listen host, invalid port, scoped Allow
- Registry integration tests: non-local publish with basic auth, temp file-backed storage, static streaming, batch POST

**Docs:**
- README documenting opt-in contract and known Express/Fastify differences


ricardo-devis-agullo:
@copilot build failed. check why and try to fix

